PR #111, opened 2020-09-23
Heuristics

martinballa:
Added back the heuristics for OSLA, MCTS and RHEA. 
To use a heuristic simply just pass it in the agent's constructor:
`players.add(new OSLAPlayer(new PandemicHeuristics());`
In the rollouts there are checks if the agent has a heuristic assigned or not. If there is one it uses that for evaluating the game state if not it uses the game's getScore function.

### DIFF
--- a/src/main/java/core/AbstractPlayer.java
+++ b/src/main/java/core/AbstractPlayer.java
@@ -1,6 +1,5 @@
 package core;
 
-
 import core.actions.AbstractAction;
 
 public abstract class AbstractPlayer {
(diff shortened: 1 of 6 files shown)


rdgain:
I have a little comment, can we keep the default constructors as well (no arguments)? So there's a very simple way to create agents with default settings, in this case keeping the heuristic null.

martinballa:
I kept the default constructor as they were, I've just added new ones.

rdgain:
Oh, they didn't show up in the summary. Awesome, good to merge! :)

rdgain (approved):

PR #115, opened 2020-10-25
Parameter search with NTBEA

hopshackle:
Integration of NTBEA to allow MCTS parameters to be optimised neatly for any game. (RMHC to follow)
(Note change to pom.xml to add dependency on com.github.hopshackle:NTBEA)

ParemeterSearch.java is the main entry point. Command line arguments are:
```
The first three arguments must be 
	<filename for searchSpace definition>
	<number of NTBEA iterations>
	<game type>Then there are a number of optional arguments:
	nPlayers=      The total number of players in each game (the default is game.Min#players) 
 	evalGames=     The number of games to run with the best predicted setting to estimate its true value (default is 20% of NTBEA iterations) 
	opponent=      The filename for the agent used as opponent. Default is a Random player. 
	useThreeTuples If specified then we use 3-tuples as well as 1-, 2- and N-tuples 
	kExplore=      The k to use in NTBEA - defaults to 100.0 
	hood=          The size of neighbourhood to look at in NTBEA. Default is min(50, |searchSpace|/100) 
	repeat=        The number of times NTBEA should be re-run, to find a single best recommendation 
	verbose        Will log the results marginalised to each dimension, and the Top 10 best tuples for each run 
```
A JSON file specifies the searchSpace, for example below. Only parameters linked to a JSONArray will be searched. The single values are defaults.
```
{
"algorithm":"MCTS",
"K" : [0.01, 0.03, 0.1, 0.3, 1.0, 3.0, 10.0, 30.0, 100.0, 300.0],
"rolloutLength" : [5, 10, 20, 50, 100, 200, 300, 500],
"rolloutsEnabled" : [false, true],
"rolloutType" : "Random",
"budgetType" : 2,
"budget" : 4000,
"epsilon" : 1e-6
}
```

Additionally, a specific opponent can be detailed in a separate JSON file. (Same format as above, but obviously don't use any JSONArrays.)
An OSLA agent can be specified by:
```
{
"algorithm":"OSLA",
}
```

And a hand-coded agent with a no-parameter constructor using:
```
{
"algorithm":"heuristic",
"class" : "games.dominion.BigMoney"
}
```


src/main/java/evaluation/ParameterSearch.java
```diff
@@ -0,0 +1,210 @@
...
+            throw new AssertionError("Must specify at least three parameters: searchSpace, NTBEA iterations, game");
+        String searchSpaceFile = args[0];
+        int iterationsPerRun = Integer.parseInt(args[1]);
+        GameType game = GameType.valueOf(args[2]);
+//        if (game != GameType.Dominion)
+//            throw new AssertionError("Only Dominion currently supported");
+        int repeats = getArg(args, "repeat", 1);
+        int evalGames = getArg(args, "evalGames", iterationsPerRun / 5);
+        double kExplore = getArg(args, "kExplore", 100.0);
+        String opponentFile = getArg(args, "opponent", "");
```
rdgain, line 49: I wonder if we can have a different argument instead here, as list of players, where each is a string of either the sample players, or a custom class name. Having a whole separate file defining 1 or 2 lines that are just strings really could be avoided in this case I think? 

Or, instead of player name/class, a JSON file could be specified (different parsing for string ending with ".json") which would also contain parameter definition for the players? That would make sense to be in a separate file, and allow flexibility on setting up the evaluation.

src/main/java/players/mcts/MCTSParams.java
```diff
@@ -1,19 +1,23 @@
...
+import java.io.FileReader;
+import java.util.*;
 
 public class MCTSParams extends PlayerParameters implements ITunableParameters {
 
     public double K = Math.sqrt(2);
     public int rolloutLength = 10;
     public boolean rolloutsEnabled = false;
     public double epsilon = 1e-6;
+    public String rolloutType = "Random";
```
rdgain, line 20: I think this should be a constant and not a string, to avoid misspelling.

src/main/java/players/mcts/MCTSPlayer.java
```diff
@@ -58,8 +65,17 @@ public AbstractAction getAction(AbstractGameState gameState) {
...
         root.mctsSearch();
+        if (debug)
+            System.out.println(root.toString());
+
 
         // Return best action
         return allActions.get(root.mostVisitedAction());
     }
+
+    public void setName(String name) {this.name = name;}
```
rdgain, line 76: Can this be moved to the super class?

src/main/java/evaluation/ParameterSearch.java
```diff
@@ -0,0 +1,210 @@
...
+        boolean verbose = Arrays.asList(args).contains("verbose");
+        int nPlayers = getArg(args, "nPlayers", game.getMinPlayers());
+
+        MCTSSearchSpace searchSpace = new MCTSSearchSpace(searchSpaceFile);
```
rdgain, line 53: Something to consider here is that games also have parameter classes that can be tuned. So changing the concepts above of "opponents" to "players" and adding parsing of games parameters as well. Although I'm not sure how much extra work / classes it'd involve to add all the games in this way, maybe reworking ITunableParameters and having a different version working with that format would be enough, keeping this separate for agent optimisation.

src/main/java/evaluation/ParameterSearch.java
```diff
@@ -0,0 +1,210 @@
...
+                    .collect(Collectors.joining(", "));
+            System.out.printf("%20s has %d values %s%n", searchSpace.name(i), searchSpace.nValues(i), allValues);
+        }
+
+        NTupleSystem landscapeModel = new NTupleSystem(searchSpace);
+        landscapeModel.setUse3Tuple(useThreeTuples);
+        landscapeModel.addTuples();
+
+        NTupleBanditEA searchFramework = new NTupleBanditEA(landscapeModel, kExplore, hood);
+        long seed = 42;
```
rdgain, line 88: Hard-coded?

rdgain (changes requested):
I like! And some more general comments:
- For usability, it's good to keep things as simple as possible, but with the option for further customisation (that can also be done simply, or with full control of what's happening). Something not mentioned in the rest of the comments along these same lines is having the option to simply specify values for the variables directly (i.e. not through program arguments) - which can be done easily enough at the moment, though the condition checking the number of arguments would fail.
- Also for usability, we like having comments throughout (especially for each function to detail what it does, what arguments it takes and what it returns) :)

Something unrelated: for games we have the GameType class that keeps track of all the games, their properties and also acts as a factory for each game. I wonder if something like that could be added for the players (i.e. extending the PlayerFactory class you've added for more general usage). Maybe not for this PR, since it'd take a bit of fiddling with how players are read by the game as well, but to add to future to-dos.

src/main/java/evaluation/GameEvaluator.java
```diff
@@ -0,0 +1,91 @@
...
+        List<AbstractPlayer> allPlayers = new ArrayList<>(nPlayers);
+
+        // We can reduce variance here by cycling the playerIndex on each iteration
+        int playerIndex = nEvals % nPlayers;
+        for (int i = 0; i < nPlayers; i++) {
+            if (i != playerIndex) {
+                int oppIndex;
+                do {
+                    oppIndex = rnd.nextInt(opponents.size());
+                } while (avoidOppDupes && allPlayers.contains(opponents.get(oppIndex)));
```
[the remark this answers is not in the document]
hopshackle: Urgent performance fix added :-)

src/main/java/core/AbstractPlayer.java
```diff
@@ -18,6 +18,7 @@
     public final int getPlayerID() {
         return playerID;
     }
+    public void setPlayerID(int id) {playerID = id;}
```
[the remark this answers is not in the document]
hopshackle: I've backed this out, so that playerID is purely package-level access.

While the only rollout strategy supported is Random, then this is not actually needed (but it is needed for a decent Dominion player). I'll raise a separate issue to deal with the issue around playerID property of an AbstractPlayer.

I personally like the idea of having a Map<Integer, AbstractPlayer> in Game (or perhaps better, AbstractGameState). In terms of opponent modelling, when rolling a state forward, then we:
i) State.copy(ID of perspective player);
ii) State.setOpponentModels(ID of opponent, AbstractPlayer)

and then roll forward. Anyway. Need to discuss this properly.

src/main/java/evaluation/ParameterSearch.java
hopshackle: Done. ParameterSearch and PlayerFactory now support a short-hand notation as (I think) yous suggested. 
At the moment the code only supports a single opponent; so this is straightforward. I do like the idea of having a single config file for more complex experiments/runs that details all the game, player, opponent parameters in one place. But..future enhancement?

src/main/java/players/mcts/MCTSParams.java
hopshackle: Yup. done.

src/main/java/players/mcts/MCTSPlayer.java
hopshackle: Done.

src/main/java/evaluation/ParameterSearch.java
hopshackle: In principle, a big "yes". Any objections if we leave that to a separate Issue/Pull Request?

I've raised an issue already to properly integrate ITunableParameters with the NTBEA libraries.

src/main/java/players/mcts/MCTSParams.java
hopshackle: At the moment we only have one possible value, and no way to change ths default :-)

On that basis, can we leave this to later?

src/main/java/evaluation/ParameterSearch.java
hopshackle: Done. Now nice and soft.

src/main/java/players/mcts/MCTSParams.java
rdgain: But it's ugly!

hopshackle: All done in the next commit...

src/main/java/players/mcts/MCTSParams.java
```diff
@@ -1,99 +1,58 @@
...
 
+    public MCTSParams() {
+        this(System.currentTimeMillis());
+    }
     public MCTSParams(long seed) {
         super(seed);
+        addTunableParameter("K", Math.sqrt(2), Arrays.asList(0.0, 0.1, 1.0, Math.sqrt(2), 3.0, 10.0));
+        addTunableParameter("rolloutLength", 10, Arrays.asList(6, 8, 10, 12, 20));
+        addTunableParameter("rolloutsEnabled", false, Arrays.asList(false, true));
+        addTunableParameter("epsilon", 1e-6);
```
rdgain, line 26: It seems a bit repetitive adding these (and misleading since the method is called addTunableParameter, but this is not tunable). I wonder if there's an easy way to get around it... (a note for the future)

src/main/java/players/mcts/MCTSSearchSpace.java
```diff
@@ -0,0 +1,135 @@
+/*
```
rdgain, line 1: Ugly to have a class that's all commented out, maybe keep it local and put it back in the branch after merge if needed?

src/main/java/games/GameType.java
```diff
@@ -117,11 +186,12 @@ public GameType stringToGameType(String game) {
...
      * Creates an instance of the given game type, with a specific number of players and game seed.
      * Add here all games implemented.
+     *
      * @param nPlayers - number of players taking part in the game, used for initialisation.
-     * @param seed - seed for this game.
+     * @param seed     - seed for this game.
      * @return - instance of Game object; null if game not implemented.
      */
-    public Game createGameInstance(int nPlayers, long seed) {
+    public Game createGameInstance(int nPlayers, long seed, AbstractParameters params) {
```
rdgain, line 194: Since we're passing params here, should probably use it within the method - potentially checking if null to create new default params (as currently), otherwise use the ones passed?

rdgain (changes requested):
The forward model next() thing we were talking about isn't included in this right? (Just checking if I missed it)

Yes - that's correct.

src/main/java/players/mcts/MCTSParams.java
hopshackle: I played with a few options here. Initially I tried reducing boilerplate by not having these as local properties, but that meant any use of a parameter had to call params.getParameterValue(name)...which was horrendous, and self-defeating.

Two specific points:
1) Parameters with a default value but no list of options **are** still tunable, so I don't think the method is ill-named. A custom-JSON file can still be used to specify a range of epsilon if you really wanted to.

2) This is still a significant reduction in boilerplate code compared to the previous version, in which default values were manually set up in code at one point, and possible values set up at another. I'm not saying this is perfect - but it feels more natural to me to set up all the parameter details in a single place in the constructor (plus, I admit, add them to the _reset() method if you want consumers to access them via a local public property).

(Open to ideas though!)

src/main/java/players/mcts/MCTSSearchSpace.java
hopshackle: Oops. That was my mistake. I just forgot to delete it once I was happy I hadn't missed anything.

src/main/java/games/GameType.java
hopshackle: Hmm....I definitely recall doing exactly that. I seem to have lost some code snippets somewhere along the way. I shall resurrect them.

Good spot!

src/main/java/players/mcts/MCTSParams.java
rdgain: Yeah those are good points. Not sure what would be better, maybe we'll find something else later.

rdgain (approved):
I think ready to merge!

Issue #120, opened 2020-11-05
Java 11 needed


erfont:
Hi, 
the project has a dependency in SolutionEvaluator that needs Java 11 at least. Trying to build it with Java 8 got me this error:

java: cannot access evodef.SolutionEvaluator
  bad class file: .../.m2/repository/com/github/hopshackle/NTBEA/0.2/NTBEA-0.2.jar!/evodef/SolutionEvaluator.class
    class file has wrong version 55.0, should be 52.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.  

Please update that info on the setting up section. 

Otherwise really great framework. Congrats!

hopshackle:
Thanks for pointing that out!

We'll get that sorted (ideally by amending the problematic library dependency).

hopshackle:
Dependency now removed. The whole framework should be be restored to Java 8 compatibility, including its transitive dependencies.

PR #121, opened 2020-11-05
Amended to use 0.2.1 of NTBEA for Java 8 compatibility

hopshackle:
Currently the 0.2 library of NTBEA is compiled under Java 11, and so requires this in ModernBoardGames.

A new version 0.2.1 is compiled under Java 8 to removed this unwanted dependency. The ModernBoardGames pom.xml file is now updated to use 0.2.1 as a dependency.


rdgain (approved):
Tested with Java 8 and project now rebuilds successfully after getting the same error reported in issue #120 .

PR #122, opened 2020-11-10
Open Loop, Information Set and RegretMatching MCTS (amongst others)

hopshackle:
Major change here is a set of MCTS enhancements to implement various algorithmic variants from the literature. This has involved some significant 'refactoring' of SingleTreeNode in particular. The major design point is that child nodes are no longer kept in an array with a fixed order, but in a Map keyed by the AbstractAction that leads to the node.

MCTSParams has also had a fair amount added to it:
1) SelectionPolicy can be ROBUST or SIMPLE. ROBUST (the old default) picks the final action based on the number of visits, while SIMPLE picks the action with the highest mean value.

2) RolloutsEnabled has been subsumed into RolloutLength. RolloutLength ==0 implies no rollouts.

3) OpenLoop can be FALSE or TRUE. FALSE is the old default, which assumes deterministic environments and does not advance the GameState as the Tree is descended. TRUE now supports OpenLoop MCTS, in which the game state is advanced on every iteration as we extend the tree. (For this reason I have added a copyCount as well as fmCallsCount, and for the moment set the fmCallsBudget to apply to the sum of the two; the old code was advancing fmCallsCount as the tree was descended, even though the FM was never called.)

4) OpponentTreePolicy has 3 values SelfOnly, Paranoid, Paranoid2. SelfOnly means that we ignore our opponents in the Tree, and only construct this using our actions. Paranoid2 is the original version - assuming that opponents are attempting to minimise our score. Paranoid is my version of this - I suspect, subject to feedback/review - that the Exploration bonus was misapplied in Paranoid2.

5) I have split RolloutLength into two parts: MaxTreeDepth and RolloutLength. MaxTreeDepth limits the depth of the tree (you'd never have guessed!), while RolloutLength now applies purely to the rollout outside the tree. This does mean that currently you cannot restrain  the sum of the two to a specific value - but it does mean that Monte Carlo Search is now supported with MaxTreeDepth=1, RolloutLength = High.

6) Redeterminise can be TRUE or FALSE. (FALSE is the old functionality). TRUE is a very simple form of Information Set MCTS that redeterminised the root game state used for each iteration. We should expect this to be significantly better on stochastic games.

7) TreePolicy can be any of UCB, AlphaGo, EXP3, RegretMatching. UCB is the standard logic, AlphaGo replaces the form of the UCB exploration term with that used by AlphaGo (instead of sqrt(logN / n) it uses sqrt(N) / (1 + n)).
EXP3 and Regret Matching implement the bandit algorithms of the same names instead of UCB. These have an exploration term, which is configured using exploreEpsilon as a new parameter (this is not used outside of these variants).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,5 @@ target/
 target/classes/META-INF/ModernBoardGame.kotlin_module
 
 out/production/ModernBoardGame/META-INF/
-*.kotlin_module
-target/classes/META-INF/ModernBoardGame.kotlin_module
 
 javadoc/
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,7 @@
         <dependency>
             <groupId>com.github.hopshackle</groupId>
             <artifactId>NTBEA</artifactId>
-            <version>0.2</version>
+            <version>0.2.1</version>
         </dependency>
     </dependencies>
 
--- a/resources/manifests/ParameterSearch/META-INF/MANIFEST.MF
+++ b/resources/manifests/ParameterSearch/META-INF/MANIFEST.MF
@@ -1,0 +1,3 @@
+Manifest-Version: 1.0
+Main-Class: evaluation.ParameterSearch
+
--- a/resources/manifests/ParameterSearch/META-INF/MANIFEST.MF
+++ b/resources/manifests/ParameterSearch/META-INF/MANIFEST.MF
@@ -1,3 +1,0 @@
-Manifest-Version: 1.0
-Main-Class: evaluation.ParameterSearch
-
--- a/src/main/java/core/AbstractGameState.java
+++ b/src/main/java/core/AbstractGameState.java
@@ -126,7 +126,7 @@ public abstract class AbstractGameState {
      * @param playerId - player observing the state
      * @return - reduced copy of the game state.
      */
-    final AbstractGameState copy(int playerId) {
+    public final AbstractGameState copy(int playerId) {
         AbstractGameState s = _copy(playerId);
         // Copy super class things
         s.turnOrder = turnOrder.copy();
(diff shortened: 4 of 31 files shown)


hopshackle:
ColtExpress bug fix: 
When the state was copied from a player's perspective using copy(playerID), there was an occasional problem with a bullet card being put into the stack of planned actions to replace an unseen card. This has now been fixed.

hopshackle:
Exploding Kittens bug fix:
There was an edge case on a Favor action that caused an empty action list to be returned by fm._computeAvailableActions(). This occured when the target had no cards left in hand by the time the Favor triggered. This could occur for example when the target had cards when the Favor was played, but a sequence of Nopes meant it was empty by the time they had to GiveCards. In this case a GiveCard action is now provided with a target of -1 (interpreted as no cards to give..so just move on).

hopshackle:
Virus bug fix. A small change to the forward model to prevent duplicate actions being created. For example with two Wild Virus cards in hand, two actions would be generated for each opponent card (this is technically all fine, but the new MCTS changes that key a map on AbstractAction was a little distressed). This also helps any algorithm by removing spurious duplicates from consideration.

hopshackle:
> Given that there are so many changes to MCTS, can we keep this as a separate algorithm and also leave the simpler version in? For those new to AI/Java/maybe even the framework, it would be useful to have a simple version in that is easier to understand. 

@rdgain 
I was all set to do this, and then on review of the code I had second thoughts - as actually I don't think the code changes are that major, or essentially more difficult to understand (I admit to being biased though). And having two (plus) separate MCTS algorithms could be more confusing? It could definitely be more bug-prone if we have the same code implemented in multiple places. 
(I have now amended MCTS and PlayerParams to default to the original settings.)
Let's talk this over?

PR #126, opened 2020-11-20
Available action refactor + addition of history to GameState

hopshackle:
This stems from a discussion/brainstorm with @rdgain and @martinballa.

The main idea is to avoid storing the AvailableActions on the AbstractGameState, and instead use ForwardModel to calculate these when needed. This was on the basis that the actions are not a core part of GameState, and this was just being used as a storage point with a pattern of:
- ForwardModel._computeAvailableActions(GameState) 
- player.getAction(GameState)

The first step is essential, but its effect is not directly visible. The new set up is:
- actions = ForwardModel._computeAvailableActions(GameState) 
- player.getAction(GameState, actions)

This makes the dependency explicit - and also allows situations where we want to restrict the acitons available for some reason.

The second change is to add a history of all actions taken to the GameState. This has also been added to the AbstractGUI as a new panel next to the GameInfo panel at the top of the screen. The idea is that this is tremendously helpful for debugging.

### DIFF
--- a/src/main/java/core/AbstractGUI.java
+++ b/src/main/java/core/AbstractGUI.java
@@ -10,6 +10,7 @@ import java.awt.*;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.stream.Collectors;
 
 @SuppressWarnings("rawtypes")
 public abstract class AbstractGUI extends JFrame {
@@ -24,6 +25,9 @@ public abstract class AbstractGUI extends JFrame {
     protected int maxActionSpace;
     protected ActionController ac;
     protected JLabel gameStatus, playerStatus, turnOwner, turn, currentPlayer, gamePhase;
+    protected JTextPane historyInfo;
+    protected JScrollPane historyContainer;
+    private int actionsAtLastUpdate;
     private WindowInput wi;
 
     public AbstractGUI(ActionController ac, int maxActionSpace) {
@@ -35,6 +39,7 @@ public abstract class AbstractGUI extends JFrame {
         turnOwner = new JLabel();
         turn = new JLabel();
         currentPlayer = new JLabel();
+        historyInfo = new JTextPane();
 
         this.wi = new WindowInput();
         addWindowListener(wi);
@@ -134,11 +139,16 @@ public abstract class AbstractGUI extends JFrame {
         gameInfo.add(turn);
         gameInfo.add(currentPlayer);
 
-        gameInfo.setPreferredSize(new Dimension(width, height));
+        gameInfo.setPreferredSize(new Dimension(width/2 - 10, height));
 
         JPanel wrapper = new JPanel();
+        wrapper.setLayout(new FlowLayout());
         wrapper.add(gameInfo);
-        wrapper.setLayout(new GridBagLayout());
+
+        historyInfo.setPreferredSize(new Dimension(width/2 - 10, height));
+        historyContainer = new JScrollPane(historyInfo);
+        historyContainer.setPreferredSize(new Dimension(width/2 - 25, height));
+        wrapper.add(historyContainer);
         return wrapper;
     }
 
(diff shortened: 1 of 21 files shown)


src/main/java/core/AbstractGUI.java
```diff
@@ -66,7 +71,7 @@ protected void setFrameProperties() {
      */
     protected void updateActionButtons(AbstractPlayer player, AbstractGameState gameState) {
         if (gameState.gameStatus == Utils.GameResult.GAME_ONGOING) {
-            List<AbstractAction> actions = gameState.getActions();
+            List<AbstractAction> actions = player.forwardModel.computeAvailableActions(gameState);
```
rdgain, line 74: I think we should pass the actions into this method to avoid slowing down the GUI with extra computations. The call to this method can be traced up to the main loop in the Game class where we do have the actions calculated - it'll take a couple method signatures to be changed, but I think it might be better that way? (similarly to how they're passed to the players)

src/main/java/core/AbstractGameState.java
```diff
@@ -223,6 +215,26 @@ public final double getScore(int playerId) {
...
     }
 
+    /**
+     * Used by ForwardModel.next() to log history (very useful for debugging)
+     *
+     * @param action The action that has just been applied (or is about to be applied) to the game state
+     */
+    protected void recordAction(AbstractAction action) {
+        history.add(action);
+        historyText.add("Player " + this.getCurrentPlayer() + " : " + action.getString(this));
```
rdgain, line 225: I wonder if it's worth storing just the action and player ID in the history? And then compute text on the fly as needed, to avoid making the game state too heavy. Although I don't know how much  this influences, and we'd need the game states for the history to print out the actions too... Mm thoughts?

src/main/java/core/AbstractPlayer.java
```diff
@@ -30,6 +32,7 @@ public final int getPlayerID() {
     public final AbstractForwardModel getForwardModel() {
         return forwardModel;
     }
+    public final void setForwardModel(AbstractForwardModel fm) {forwardModel = fm;}
```
rdgain, line 35: Is this used anywhere? This was avoided for a similar reason as the player ID, to avoid having those settable from outside the package and let the core code handle things correctly.

src/main/java/games/tictactoe/gui/TicTacToeGUI.java
```diff
@@ -52,7 +52,7 @@ public TicTacToeGUI(Game game, ActionController ac) {
     @Override
     protected void updateActionButtons(AbstractPlayer player, AbstractGameState gameState) {
         if (gameState.getGameStatus() == Utils.GameResult.GAME_ONGOING) {
-            List<AbstractAction> actions = gameState.getActions();
+            List<AbstractAction> actions = player.getForwardModel().computeAvailableActions(gameState);
```
rdgain, line 55: Same as above with GUI update receiving the actions.

src/main/java/games/pandemic/gui/PandemicGUI.java
```diff
@@ -311,7 +311,7 @@ protected void clearAllHighlights() {
 
     protected void updateActionButtons(AbstractPlayer player, AbstractGameState gameState) {
         int id = player.getPlayerID();
-        List<AbstractAction> actions = gameState.getActions();
```
rdgain, line 314: Same as above with GUI update receiving the actions.

rdgain (changes requested):
A couple small comments.

src/main/java/core/AbstractGUI.java
hopshackle: Possible - but on this one I disagree philosophically :-). This is GUI code, so speed seems irrelevant; the time between GUI tick updates is in geological time compared to any computation of available actions, so I didn't even think about optimising this.
I may have missed something though?

rdgain: I like my GUIs speedy! But that's fine 👍

src/main/java/core/AbstractGameState.java
hopshackle: Yup - I thought about this too.  The implementation *should* be highly efficient - the key point is that each String is only added once and never, ever copied. The two crucial code optimisations are:
1) in AbstractGameState.copy(), where we explicitly do not copy actions (just their references), or their string representations on the basis that these are now immutable history. This means that the same String is used in all copied future states.
        s.history = new ArrayList<>(history);
        s.historyText = new ArrayList<>(historyText);
            // we do not copy individual actions in history, as these are now dead and should not change

2) In AbstractGameState.equals():
        // we deliberately exclude history from this equality check

So - I think that the memory and processing footprint of this is actually quite slim. (If one deliberately copied all those Strings and Actions each time though....)

src/main/java/core/AbstractPlayer.java
hopshackle: No - it's not used anywhere - I'll put this back to package-view.

(I think I may have used it on a test case at some point, and then decided against. I can also see future developments that might use this; if for example we have a player that learns a forward model, or sets up an opponent model that uses a simplified one...but that's way in the future.)

rdgain: Yeah I can see that happening, but we can deal with it at that point :D

rdgain (approved):

PR #129, opened 2020-12-23
Diamant

montoliu:


src/main/java/games/diamant/actions/ContinueInCave.java
```diff
@@ -0,0 +1,35 @@
...
+    @Override
+    public boolean execute(AbstractGameState gs) {
+        // Nothing to be executed. The actions are executed in the ForwardModel
+        return true;
+    }
+
+    @Override
+    public AbstractAction copy() {
+        return new ContinueInCave();
+    }
```
hopshackle, line 17: even easier you could just return this from copy() given lack of any mutable state

src/main/java/games/diamant/DiamantGameState.java
```diff
@@ -0,0 +1,253 @@
...
+import java.util.*;
+
+import static core.CoreConstants.PARTIAL_OBSERVABLE;
+
+public class DiamantGameState extends AbstractGameState implements IPrintable {
+    Deck<DiamantCard>          mainDeck;
+    Deck<DiamantCard>          discardDeck;
+    Deck<DiamantCard>          path;
+
+    List<DiamantTreasureChest> treasureChests;
```
hopshackle, line 26: Following on from previous comment, is there any difference (apart from name) between DiamantHands and TreasureChests?

src/main/java/games/diamant/DiamantForwardModel.java
```diff
@@ -0,0 +1,313 @@
...
+    {
+        return new DiamantForwardModel();
+    }
+
+    /**
+     * Play the card
+     * @param card: card to be played
+     * @param dgs: current game state
+     */
+    private void playCard(DiamantCard card, DiamantGameState dgs)
```
hopshackle, line 267: All three places that call fm.playCard() have the same two lines immediately preceding:
            DiamantCard card = dgs.mainDeck.draw();
            dgs.path.add(card);
Could these be added into playCard() - removing one parameter, and reducing code duplication?

hopshackle (approved):
All looks good - I have a few points of style; but that's all.

src/main/java/games/diamant/DiamantHeuristic.java
```diff
@@ -0,0 +1,42 @@
...
+                    max_ngens = nGems;
+
+            }
+        }
+
+        int player_gems = dgs.treasureChests.get(playerId).GetNumberGems();
+        double score;
+        if      (player_gems > max_ngens) score = 1.0;
+        else if (player_gems < max_ngens) score = -1.0;
+        else                              score = 0.0;
```
hopshackle, line 39: Technically I believe we should only hit +/-1.0 in the case of a Win/Lose condition. Before the end of the game, the returned score should be of lower magnitude...easiest might be just to have +/- 0.5 in this case. Or, return some direct function of player_gems.

Also - player_gems > max_ngens will always be false here? Do you mean:
if      (player_gems == max_ngens) score = 1.0;

PR #130, opened 2020-12-28
DominionCoreChanges

hopshackle:
A number of changes originally implemented in the Dominion branch, but which are not formally part of the game.
Hence I've split them out as a separate Pull Request to make it a bit easier to review - these need more review given their core impact.

The main headline changes are:

1) Heuristic functions are now all set up as TunableParameters. This means they can be tuned as part of MCTS or RMHC parameter tuning via JSON definition files.
Documentation to be added tp the wiki shortly.
As part of this, the heuristic used in MCTS is now defined as part of MCTSParams, rather than an MCTSPlayer constructor.
All existing Heuristics have been converted to the new format (and one has been added for DotsAndBoxes).

2) getGameScore() added as a new abstract method to Game. The point of this is to provide the current game score independent of any heuristics.
For some games (TicTacToe, Exploding Kittens for example) this makes no sense and can be (and has been) set to default to 0. 
However in games like ColtExpress, Dominion, LoveLetter, Uno there is a very natural sense of 'current score' (loot, victory points, affection tokens, card values of opponents respectively).
This has also been added to the GameStateInfoPanel in the GUI.
[The point here is also to support research into what happens under three very natural, but different, objective functions: "Win/Lose", "Ordinal posiition", "Score".]

3) Added some hooks to support full game event/result logging. GameEvents (defined in CoreConstants) will trigger for the end of a game/turn/round, and each time a decision is taken by an agent.
IGameListeners can be registered to listen to these events for any purpose. 

4) Then, linked to point 3) above, added GameLogger/IGameAttribute as one possible use for these Events to generate detailed csv files of game results for later statistical/other analysis.

5) Component.hashCode is no longer final. I have however made it final on all core inherited sub-classes. This now means that Components that are naturally composite can override equals/hashCode together. 
This affects Deck, GraphBoard, GridBoard and Area, which now include the componentIDs of their contents in equals and hashcode. 
This will cause problems if a Component with mutable state is used as a HashMap key - but that should be a silly thing to do, right?

6) Added some Deck utility methods to play nicely with streams.


src/main/java/players/rmhc/RMHCPlayer.java
```diff
@@ -21,6 +21,7 @@
     private double avgTimeTaken = 0, acumTimeTaken = 0;
     private int numIters = 0;
     private int fmCalls = 0;
+    private int copyCalls =0;
```
rdgain, line 24: This is not part of what we've traditionally used as "FM call budget", specifically it's the number of calls to the FM.next() function. We can add another type of budget to include copies as well, but it should not affect the other settings for comparison to previous experiments.

src/main/java/players/rmhc/RMHCPlayer.java
```diff
@@ -106,10 +111,6 @@ private void runIteration(AbstractGameState stateObs) {
         avgTimeTaken = acumTimeTaken / numIters;
     }
 
-    @Override
-    public String toString() {
```
rdgain, line 110: Why remove?

src/main/java/players/mcts/MCTSEnums.java
```diff
@@ -7,7 +7,7 @@
 public class MCTSEnums {
 
     public enum strategies {
-        RANDOM
+        RANDOM, Dominion_BigMoney, Dominion_PlayActions
```
rdgain, line 10: This is a strange thing to see in a general player. Can these be in a separate enum if needed for some experiments (or even a Dominion-specific MCTS player)?

src/main/java/games/coltexpress/ColtExpressGameState.java
```diff
@@ -199,10 +199,15 @@ protected AbstractGameState _copy(int playerId) {
...
 
     @Override
-    protected double _getScore(int playerId) {
+    protected double _getHeuristicScore(int playerId) {
         return new ColtExpressHeuristic().evaluateState(this, playerId);
     }
 
+    @Override
+    public double getGameScore(int playerId) {
+        return getLoot(playerId).sumInt(Loot::getValue);
```
rdgain, line 208: (This comment is repeated in several places) - reflection in Java is very slow, and to be avoided if at all possible. It makes code look prettier, but we need it more to be fast, so I'd highly recommend changing all these to use old-school references to objects.

src/main/java/core/Game.java
```diff
@@ -677,16 +681,34 @@ public static void main(String[] args) {
         /* 3. Set up players for the game */
         ArrayList<AbstractPlayer> players = new ArrayList<>();
 
-//        players.add(new RandomPlayer());
-        players.add(new RMHCPlayer());
-        players.add(new MCTSPlayer());
-        players.add(new OSLAPlayer());
+        MCTSParams params1 = new MCTSParams();
```
rdgain, line 684: I'm guessing this is for some experiments, but it makes the main entry point to running the framework complex and foreign to someone just getting started with spelled-out parameters for MCTS. Can we keep this main method super small and tidy, and maybe have other run classes/methods that do fancier things?

src/main/java/core/Game.java
```diff
@@ -94,25 +98,6 @@ public final void reset() {
         resetStats();
     }
 
-    /**
-     * Resets the game. Sets up the game state to the initial state as described by game rules,
-     * and initialises all players.
-     *
-     * @param newRandomSeed - random seed is updated in the game parameters object and used throughout the game.
-     */
-    public final void reset(long newRandomSeed) {
```
rdgain, line 103: I didn't check the full class, but this is not needed anymore?

src/main/java/core/AbstractPlayer.java
```diff
@@ -24,6 +24,7 @@
     public final int getPlayerID() {
         return playerID;
     }
+    public void setPlayerID(int id) {playerID = id;}
```
rdgain, line 27: Is there a need for this again?

src/main/java/core/AbstractGameState.java
```diff
@@ -202,7 +221,7 @@ public final AbstractGameState copy() {
      * @return - double, score of current state.
      */
     public final double getScore(int playerId) {
-        return _getScore(playerId);
+        return _getHeuristicScore(playerId);
```
rdgain, line 224: A problem here. It's good to split the heuristic and game score, yes. But a function named "getScore" becomes ambiguous then, and given this split I'd actually expect it to return the game score.

src/main/java/core/AbstractGameState.java
```diff
@@ -163,7 +172,17 @@ public final AbstractGameState copy(int playerId) {
...
+
+    /**
+     * This provides the current score in game turns. This will only be relevant for games that have the concept
+     * of victory points, etc.
+     * If a game does not support this directly, then just return 0.0
+     *
+     * @param playerId
+     * @return - double, score of current state
+     */
+    public abstract double getGameScore(int playerId);
```
rdgain, line 185: I'm wondering if there are more games that have a scoring systems than those that don't. If so, having this method abstract and forcing all games to implement it is fine. But forcing all games to implement it just to return 0 in a majority is not ideal, and could easily just be done in the core code (and any other games that do have scoring can overwrite it). Maybe something to revise at some point.

rdgain (commented):
Tiny thing: the Dominion card back is not a core change.

There's also a comment on reflection that claims it's repeated in several places, but I missed all the other places where I wanted to add it so it's actually in only one place. That's probably the biggest thing in my comments, otherwise looks like good work to polish things up, thanks!

src/main/java/games/coltexpress/ColtExpressGameState.java
hopshackle: I've looked into this in rather more detail, and run some experiments documented at https://github.com/GAIGResearch/TabletopGames/wiki/Coding-Conventions

The conclusion is that I think this is absolutely fine - there is no performance difference between lambda usage and method references. However I am going to avoid streams in highly core pieces of code (see the wiki link for my full argument).

src/main/java/players/rmhc/RMHCPlayer.java
hopshackle: Very fair. I'll separate these two. In passing it migt be worth noting that for all the eight implemented games currently, the copy() method quite a bit more time-consuming then the next() method.

(My main beef with the previous code as actually that in Closed Loop MCTS it incremented the fmCalls budget even though the fm wasn't called! Removing this too naively then led to a potential infinite loop...but I've now added a specific check for that so that we stop if the number of iterations ever exceeds the Forward Model budget)

src/main/java/players/rmhc/RMHCPlayer.java
hopshackle: Because I have moved name up to the abstract super class, so the behaviour is the same. (Or at least, it would have been if I hadn't accidentally removed the line that set the default name to "RMHC"...that is now being added back in . Ahem.)

src/main/java/players/mcts/MCTSEnums.java
hopshackle: It's a fair cop. I shudder at the thought of a Dominion-specific MCTS player, but I'm sure there can be a more general method to inject game-specific rollout strategies. (And this is now actually less immediately necessary for experiments, so i'll remove these.)

src/main/java/core/Game.java
hopshackle: Agreed. Will simplify and tidy up.

src/main/java/core/Game.java
hopshackle: Nothing used it..so I aggressively deleted it as non-used cruft. The other two reset() methods were all used regularly...I can put it back if you feel it is useful.

src/main/java/core/AbstractPlayer.java
hopshackle: Indeed not...it crept back by default. And I think I now have a way round the original problem. Will remove again.

src/main/java/core/AbstractGameState.java
hopshackle: I agree. In this case I was not aggressive enough, and didn;t want to change the name of the existing method. Now I have permission I'll refactor this!

src/main/java/core/AbstractGameState.java
hopshackle: My view on this was that a score is a very common feature in games (not, admittedly always so, as Exploding Kittens shows). I felt that it was always worth asking the implementor of a game to ask themselves this question, especially as if the answer is 'No', then we have clear javadoc to tell them what to put in as a default...and it shouldn't take them more than a second or two to write "return 0.0".
Having a default implementation to do this I felt could be dangerous if future implementors don;t realise they *should* ask themslves this question.

src/main/java/games/coltexpress/ColtExpressGameState.java
rdgain: That looks like a great wiki page, good things to note, thanks! :)

rdgain (approved):

Issue #133, opened 2021-01-14
Proposed change to Deck Visibility


hopshackle:
This stems from some work to report consistent metrics across games in terms of Action Space / Branching Factor / Hidden Information level and so on. There is a fair bit of game-specific coding I had to do to report the number of components in a game state, and the percentage that are hidden (see GameReportII.java in the Dominion branch).

This is a proposal to amend the Deck (and a couple of other Component classes) to standardise this interface a little (without, I hope, imposing any further load on future game implementors; and there are some follow-on possibilities that might make things easier).

- Deck only contains the details on the individual components that make up the deck. 

- PartialObservableDeck provides information on the individual components, and on their visibility to players. is great, but also means that it has information on the number of players in a game through the length of the visibility array. This has to be injected at instantiation (implicitly).

When scanning gameState.getAllComponents(), PartialObservableDecks are easy to scan to calculate the number of objects, and visibility metrics. However the treatment of Decks is game/component dependent. There are three standard forms:
 1. Whole deck is invisible to all players (e.g. a Draw deck)
 2. Whole deck is visible to all players (e.g. a Discard deck)
 3. Whole deck is visible to its owner, and invisible to everyone else (e.g. a Player's Hand)
However, there is no information attached to the Deck to indicate which of these applies. It is possible mostly to get a count - for those games that implement this - via the _getAllUnknownComponents() method; but this does not necessarily include the same thing consistently across all games.

My proposal is to add a new interface (or just methods to Deck), on the lines of:
```java
    void setVisibilityMode(VisibilityMode visibilityStatus);

    VisibilityMode getVisibilityMode();

    public enum VisibilityMode {
        VISIBLE_TO_ALL, HIDDEN_TO_ALL, VISIBLE_TO_OWNER, ITS_COMPLICATED
    }
```
A default implementation would be for something to be ```HIDDEN_TO_ALL```, as the most common setting.

The slightly ugly element to this is that for PartialObservableDeck these two will be overridden to be ignored in favour of the existing functionality - this is the meaning of ```ITS_COMPLICATED```. A bit clunky...any better ideas?

There is then the issue that a few games have implemented their own Components that contain other Components (Compartment in ColtExpress, VirusOrgan in Virus). I was therefore thinking of adding another new interface (or perhaps adding it to the one above) with:

```java
interface ComponentContainer<T extends Component> {

    List<T> getContents(); // or make this method tie in with the current Deck implementation for consistency

}
```

[...]

PR #136, opened 2021-02-12
IComponentContainer and Heuristic tuning

hopshackle:
This contains two distinct pieces of work:

1) Implementation of Proposal #133 to standardise deck visibility across everything and make tracking levels of hidden information much easier; without requiring game implementers to do too much additional work.

2) OSLAHeuristic to enable quick and easy tuning of Heuristic agents (it's already possible to tune the Heuristic used in RMHC or MCTS, so this is filling in a gap).

### DIFF
--- a/src/main/java/core/CoreConstants.java
+++ b/src/main/java/core/CoreConstants.java
@@ -34,4 +34,8 @@ public class CoreConstants {
     public enum GameEvents {
         GAME_OVER, ROUND_OVER, TURN_OVER, ACTION_CHOSEN
     }
+
+    public enum VisibilityMode {
+        VISIBLE_TO_ALL, HIDDEN_TO_ALL, VISIBLE_TO_OWNER, ITS_COMPLICATED
+    }
 }
--- a/src/main/java/core/CoreConstants.java
+++ b/src/main/java/core/CoreConstants.java
@@ -43,6 +43,6 @@ public class CoreConstants {
      * this should cover almost all future eventualities.
      */
     public enum VisibilityMode {
-        VISIBLE_TO_ALL, HIDDEN_TO_ALL, VISIBLE_TO_OWNER, MIXED_VISIBILITY
+        VISIBLE_TO_ALL, HIDDEN_TO_ALL, VISIBLE_TO_OWNER, FIRST_VISIBLE_TO_ALL, LAST_VISIBLE_TO_ALL, MIXED_VISIBILITY
     }
 }
--- a/src/main/java/core/CoreConstants.java
+++ b/src/main/java/core/CoreConstants.java
@@ -35,7 +35,14 @@ public class CoreConstants {
         GAME_OVER, ROUND_OVER, TURN_OVER, ACTION_CHOSEN
     }
 
+    /**
+     * Used in Components that contain other Components (see IComponentContainer) to mark which players can see the
+     * contents.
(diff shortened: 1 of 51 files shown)


src/main/java/games/coltexpress/ColtExpressForwardModel.java
```diff
@@ -60,12 +61,11 @@ public void _setup(AbstractGameState firstState) {
...
             playerCards.shuffle(new Random(cep.getRandomSeed()+playerIndex));
 
-            Deck<ColtExpressCard> playerHand = new Deck<>(
-                    "playerHand" + playerIndex, playerIndex);
+            Deck<ColtExpressCard> playerHand = new Deck<>("playerHand" + playerIndex, playerIndex, VisibilityMode.VISIBLE_TO_OWNER);
 
             cegs.playerHandCards.add(playerHand);
 
-            Deck<Loot> loot = new Deck<>("playerLoot" + playerIndex, playerIndex);
+            Deck<Loot> loot = new Deck<>("playerLoot" + playerIndex, playerIndex, VisibilityMode.VISIBLE_TO_OWNER);
```
rdgain, line 68: This is actually not visible to the owner either, hidden to all. No one knows what loot values they collected until the end of the game. Although I see later that this means to distinguish between types of loot while loot values should be kept hidden. So I guess this doesn't change much the logic?

src/main/java/core/CoreConstants.java
```diff
@@ -34,4 +34,15 @@
...
+
+    /**
+     * Used in Components that contain other Components (see IComponentContainer) to mark which players can see the
+     * contents.
+     * MIXED_VISIBILITY is an indicator that none of the previous three apply, and that the IComponentContainer
+     * will need to implement more sophisticated logic. This is done for example in PartialObservableDeck - and
+     * this should cover almost all future eventualities.
+     */
+    public enum VisibilityMode {
+        VISIBLE_TO_ALL, HIDDEN_TO_ALL, VISIBLE_TO_OWNER, MIXED_VISIBILITY
```
rdgain, line 46: Some others that I'd expect to see often would be FIRST_VISIBLE (to all) or LAST_VISIBLE (to all), for the cases where you have a deck on the table that's technically face up, so you can see the first card, but you can't look through it, so the others are technically hidden. Then if a card is removed from the deck the next one on the deck would become the one visible. An immediate example of use would be in Colt Express for the round cards where visibility of the first card is set manually, but I imagine more could use this as a default implemented functionality.

rdgain (approved):
Very nice! I think things look much cleaner with code taken off the game implementations as well and hidden behind intuitive to use visibility modes. The component container also makes things clearer to understand all the stacking that's going on in a game, thanks!

src/main/java/games/coltexpress/ColtExpressForwardModel.java
hopshackle: Oh! Whenever we play Colt Express we always look at the loot tiles we pick up! Ooops. (which is why I used VISIBLE_TO_OWNER)

But...the main substantive point is that you do know the important stuff (just not which loot bags are 250 versus 500). Actually, thinking about this, this should probably be VISIBLE_TO_ALL on this logic.

src/main/java/core/CoreConstants.java
hopshackle: Nice idea!

PR #137, opened 2021-02-12
GridBoard update for framework consistency

rdgain:
GridBoard was the only component allowed to go rogue and ignore the whole component system in the framework, e.g. by placing characters/any object types in a grid, that are not necessarily components themselves, leading to strange areas conceptually. So, this PR mostly addresses the change of this class, including the removal of reflection, and keeping everything expressed in components:

- GridBoard is a generic of type <T extends Component> like Deck
- Reflection removed in GridBoard and all operations are in terms of components
- Tic Tac Toe still uses a GridBoard, but with the lowest level component instead (a Token representing each symbol). **Initially "defaultToken"s are put in the board, potentially better to remove this concept and simply place tokens with actions, to more accurately represent the game state.**
- Dots&Boxes no longer uses GridBoard, but instead some simpler data structures grouping and mapping components, for a speed up of game state copy computation and general simplification of game code
- Edges and boxes/cells in Dots&Boxes also extend Component now to keep with the consistency

### DIFF
--- a/src/main/java/core/components/GridBoard.java
+++ b/src/main/java/core/components/GridBoard.java
@@ -12,53 +12,47 @@ import utilities.Vector2D;
 
 import java.io.FileReader;
 import java.io.IOException;
-import java.lang.reflect.Array;
 import java.util.*;
 
 import static core.CoreConstants.imgHash;
 import static utilities.Utils.getNeighbourhood;
 
-public class GridBoard<T> extends Component {
+public class GridBoard<T extends Component> extends Component {
 
     private int width;  // Width of the board
     private int height;  // Height of the board
 
-    private T[][] grid;  // 2D grid representation of this board
-    private Class<T> typeParameterClass;  // Type of this class
+    private Component[][] grid;  // 2D grid representation of this board
 
     protected GridBoard() {
         super(Utils.ComponentType.BOARD);
-        typeParameterClass = (Class<T>) String.class;
     }
 
-    public GridBoard(int width, int height, Class<T> typeParameterClass) {
+    public GridBoard(int width, int height) {
         super(Utils.ComponentType.BOARD);
         this.width = width;
         this.height = height;
-        this.typeParameterClass = typeParameterClass;
-        this.grid = (T[][]) Array.newInstance(typeParameterClass, height, width);
+        this.grid = new Component[height][width];
     }
 
-    public GridBoard(int width, int height, Class<T> typeParameterClass, T defaultValue) {
-        this(width, height, typeParameterClass);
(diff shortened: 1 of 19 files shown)


src/main/java/core/components/GridBoard.java
```diff
@@ -156,7 +149,7 @@ public T getElement(int x, int y) {
      *
      * @return - 2D grid.
      */
-    public T[][] getGridValues() {
+    public Component[][] getGridValues() {
```
hopshackle, line 152: getElement() and setElement() use the more natural T as parameter/return value. Why not do this here too (as indeed was the case before the change) and return T[][]? Admittedly there is a need to cast a created array to T[][]...but we can't avoid casts to T at some point in the code.

src/main/java/core/components/GridBoard.java
```diff
@@ -178,7 +171,7 @@ public T getElement(int x, int y) {
      * @param orientation - int orientation, how many times it should be rotated clockwise
      * @return - new grid with the same elements and correct orientation.
      */
-    public T[][] rotate(int orientation) {
+    public Component[][] rotate(int orientation) {
```
hopshackle, line 174: And see previous point. As a user of GridBoard, I woul dfind it much more natural to use if it returns T[][] from all these other methods. All I'm going to do as a user is cast them to T in the client code anyway!

hopshackle (approved):
Looks good - just some comments on returning generic T[][] from GridBoard to (potentially) make it easier to use

martinballa:
Looks good to me.

src/main/java/core/components/GridBoard.java
rdgain: I would have liked that too. I prefer to stay away from reflection, which would be needed for casting the 2D array type (can't just use normal casting with array types). Since the type is known when the class is used, it's easier and faster to cast on the other side (though admittedly more work). Users can also use the get/set element methods which do cast it, so I thought it might be fine to not complicate it here? 

If not, do you know if java 8 streams or reflection would be faster? Both would be options here to return the T[][] type.

hopshackle: I've refreshed my understanding of T[] vs T casts - and in particular why casting to T[] when we know the actual array is Object[] is a poor idea (while casting to T is fine as we are confident that the actual content is a T).

While there are methods to deal with this - they might just confuse people. So I agree with the proposed solution.

Except...flattenGrid needs to be amended to return Component[] instead of T[] for consistency ;-)

rdgain: Fair point! Updated :)

Issue #134, opened 2021-02-03
Proposal to migrate IExtendedSequence into core framework


hopshackle:
In Dominion I have implemented a system for grouping a set of linked extended actions, potentially across multiple players. This is a proposal to consider migrating this into the core framework, after review by the community. This is also I think closely related to Issue #40.

There are two related use-cases that this may be helpful with:

1. Move Groups

   This is where we have to make what is formally a single 'action', but one which makes sense to break up into a set of distinct decisions to avoid combinatorial explosions. An example in Dominion is where we have to discard 3 cards out of 6. This gives a total of 20 possible actions (<sup>6</sup>C<sub>3</sub>); but might be more tractable to consider as three distinct decisions - the first card to discard (out of 6), then the second (out of 5), and then the third (out of 4). This formally gives 120 options over the three decisions, but can be very helpful where there is clearly one best card to discard, which will focus MCTS rapidly on this as the first action. An example in Catan might be first deciding to what to build (Development Card, Settlement, or Road if you have cards for any of them), and then secondly deciding where to build it.

   (The 'Move Group' comes originally (I think) from Saito et a. 2007 in Go, and the idea of first deciding whether to play a) in the corners, b) adjacent to the last piece played, c) one of the other possible board positions; and then secondly deciding which exact space to play on from the reduced options.)

2. Extended Actions

   In Dominion we can have chains of decisions that cascade from each other. For example if I play a 'Militia', each other player decides first whether to defend themselves with a Reaction card (which may in turn enable a whole set of further decision), **or** else discards down to a hand of three cards (their decision as to which to discard). This circulates round the table before I continue with my turn.

   This sort of thing is perfectly trackable within ```ForwardModel``` and ```TurnOrder``` directly (as, for example, Exploding Kittens does with a ```Nope``` action that cancels the last action card played), but when one has 15 different cards (and many more to come in expansions) that have some chain of actions in this way, it makes better design sense to try and encapsulate all this logic and tracking in one place - otherwise ```ForwardModel``` (or ```TurnOrder```, or ```GameState```) becomes very bloated.

The implementation as it stands in Dominion (and also in DiceMonastery as a work-in-progress) has the following components that I would like to consider migrating to core:

a) Stack in GameState to track the chain of interrupts and reactions (exactly like that used in ExplodingKittens):

```java
Stack<IExtendedSequence> actionsInProgress = new Stack<>();
```

[...]

PR #138, opened 2021-02-15
Moved IExtendedSequence into core framework.

hopshackle:
To implement Proposal #134.

I wrestled with a few detailed implementation options given different usage of ForwardModel._next() across games. The central issue being that when an action is executed, this information needs to be passed to the current IExtendedSequence being processed so it can update its internal state.  This code needs to **not** be in each and every implementation of ForwardModel:

```
            if (currentState.isActionInProgress()) {
                // we register the action with the currently active ActionSequence
                currentState.currentActionInProgress().registerActionTaken(currentState, action);
            }
```

My first idea was to move the central action.execute() call into AbstractForwardModel.next() ... which would have been fine for all games except ExplodingKittens and Diamant. And converting ExplodingKittens to use the new IExtendedSequence seemed like far too major and dangerous a 'refactor'. So I dropped that idea (although I still like it in principle; I believe action.execute() should be in next() and not _next()..?)

To get action.execute() into the core FM I also toyed with extending ForwardModel.next() to have two hooks that implementing ForwardModels could use, i.e.:
```
._beforeAction();
action.execute(state);
._afterAction(); // equivalent to the current _next()
```
But this didn't really solve the problem for ExplodingKittens.

I then considered putting the crucial lines of code within AbstractAction.execute() as a `public final` method - and then having ._execute() as the abstract method that all implementing sub-classes have to provide. In fact, I implemented this (see the git log for details) and let the IDE refactor the 118 sub-classes that changed :-) - I like this philosophically on the basis that registering that an action being executed feels like it should be in the execute method...but it also ran the risk of being yet another layer of OO obfuscation.

The final implementation just puts the lines into ForwardModel as the best compromise...to minimise changes to existing games. I welcome any thoughts on better approaches!


src/main/java/core/AbstractGameState.java
```diff
@@ -153,9 +162,35 @@ public final AbstractGameState copy(int playerId) {
...
+        return actionsInProgress.isEmpty() ? null : actionsInProgress.peek();
+    }
+
+    public boolean isActionInProgress() {
+        checkActionsInProgress();
+        return !actionsInProgress.empty();
+    }
+
+    public void setActionInProgress(IExtendedSequence action) {
+        if (gamePhase != DominionGameState.DominionGamePhase.Play)
```
rdgain, line 175: Dominion-specific shouldn't be in core framework.

rdgain (commented):
I think having these in either AbstractForwardModel or action.execute() would both be fine, actually leaning towards it being on the action side - with the worry that even though it's easy to refactor all the actions, git will make it an absolute pain to merge existing branches. 

In the general case all FMs execute similar steps of applying actions, checking game ends, finding the next player etc. which could all be delegated to the super class. But this takes away from the flexibility of the implementation and is not compatible with all the current games either. Actions may not be executed immediately when sent to the FM if the game rules say to wait a bit, or there may be additional conditions checked or computations to do outside the action. So I think leaving specific implementations to deal with applying the action received is fine - but for this reason and potential delays in action execution it would be better to have it on the action side when we know for sure the action has been executed. Although, I guess this also goes to the concept behind it: is this method supposed to be called when the action is played (which should be in next), or when the action was executed (which should be on the action side). I think it's actually the first case here, so the FM is the better place! (What a confusing comment...)

src/main/java/core/AbstractGameState.java
hopshackle: Indeed it should not! Thanks for catching that.

PR #140, opened 2021-03-02
Timers for agent decision making

rdgain:
- Added timer per game for agent, which is updated when agent is asked for a decision.
- Timer supports increment (time added after a decision is made)
- Timeout checks in main Game loop, which points to a function to either disqualify or play a random action for the player
- Time defined per game type (but not final, so it can be set differently for different runs), initialised in the FM.setup() (so it's reset when the game is reset)
- New timing class (ElapsedCpuChessTimer) to allow for pausing/resuming a player's clock and keeping track of overall remaining time

### DIFF
--- a/src/main/java/core/AbstractForwardModel.java
+++ b/src/main/java/core/AbstractForwardModel.java
@@ -1,6 +1,7 @@
 package core;
 
 import core.actions.AbstractAction;
+import utilities.ElapsedCpuChessTimer;
 import utilities.Utils;
 
 import java.util.Arrays;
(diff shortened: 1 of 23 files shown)


hopshackle (commented):
Looks good!

My two substantive comments are: 
1) Use GameParameters instead of GameType to store the time budgets. Seems more natural to me, and makes them more easily configurable? (But persuadable)

2) Repetition of if (TIMER) ... code in Game hurts my eyes! 

The other is really a consideration for future changes depending on what views are.

hopshackle (approved):
Yay!

Only two minor comments are:
i) As the code stands, there is a minimum increment of 1 second, or 1 minute for the game. Could the time parameters be modified to doubles to allow, say 0.5 minutes or 100ms of incremental time?

ii) MS is used as the abbreviation for Minutes in one parameter name...which might be taken for milliseconds...could this be amended to Mins?

PR #154, opened 2021-03-15
More Advanced Game Reporting

hopshackle:
This goes a little further towards generalising the Reporting facilities in TAG using GameReportII.

The idea is that we have a number of interfaces:

`IGameAttribute` : extracts a single piece of information (Attribute) from either of:
    - `(AbstractGameState state, AbstractAction action)`
    - `(AbstractGameState state, int player)`
This can be a String, Integer, Boolean, Double, or something else bespoke.

`IGameListener` : Listens to a Game (classic Observer pattern) for events that mark the start/end of game, the end of a round or an individual player's turn, or each time a player makes a decision. Generally speaking this will have a list of `IGameAttribute`s that it extracts to record interesting information about the event.

`IStatisticLogger` : This is responsible for *how* to record data (where the `IGameListener` controls *when* data is extracted, alongside the `IGameAttribute`s that define *what* is extracted. Two implementations currently either log summary data to the screen, or write detailed data to a file (one row per event). Implementations can easily be added to send data to a database, or a GUI for example.

`CoreConstants.GameEvents` define the current set of *when*: `ABOUT_TO_START, GAME_OVER, ROUND_OVER, TURN_OVER, ACTION_CHOSEN, GAME_SEQUENCE_OVER`. Default generation of these events has been added to the core framework at appropriate places, and they can be overridden as required for specific games.

`GameReportII` has then been enhanced to provide a means to tie all these together so that multiple different types of information can be generated. See the 'help' text in that class for details.
(Might it be permissible to delete the old GameReport now?)

Future work not covered in this:
 - We still have `PlayerReport` as a separate executable to provide data on Agents (reporting MCTS stats for example). IT would be nice to merge this into a single Reporting class so that we can track Agent information at the same time as Game level information (and cross-reference where useful).
 - It is necessary to instantiate `IGameListeners` for individual games at the moment. I would like to generalise this in the future so that a user can write a quick enum class extending `IGameAttribute` and then inject that into a standard set of game listeners.


src/main/java/core/CoreConstants.java
```diff
@@ -32,7 +32,7 @@
     public final static int backgroundImgHash = Hash.GetInstance().hash("backgroundImg");
 
     public enum GameEvents {
-        GAME_OVER, ROUND_OVER, TURN_OVER, ACTION_CHOSEN
+        ABOUT_TO_START, GAME_OVER, ROUND_OVER, TURN_OVER, ACTION_CHOSEN, GAME_SEQUENCE_OVER
```
rdgain, line 35: Is GAME_SEQUENCE_OVER meant to capture action sequences, or several games being played? I'm not sure that's a clear name for it.

rdgain (approved):
Only a small comment. Also, does the GUI work properly still in Tic Tac Toe with that change?

src/main/java/core/CoreConstants.java
hopshackle: Yes - just checked that TTT GUI still works.

I was also a little uncomfortable with GAME_SEQUENCE_OVER. I've amended the implementation to exclude this, so that GameEvents are actually all genuinely Game-Events. The end of a sequence of games is now controlled by a new (optional) method on the IGameListener interface.

PR #161, opened 2021-05-25
[WIP] New game: Terraforming Mars

rdgain:
This is still work in progress! The game is far along now to request playtesting to hunt for bugs (as it contains a lot of actions and variations of action types and parameters)! Happy to receive reports of any issues, but specifically looking for issues with:
- Crashes when trying to do a specific action
- Being allowed to do an action when you shouldn't be
- Not being allowed to do an action when you should be
- Visual bugs (something not correctly drawn / out of bounds etc.)
- Possible visual improvements (anything that could be better drawn / better interactions possible to make the game more intuitive / easier to play / more accessible)

Any reports along these lines with as much detail as possible to allow reproduction or intuition of the problem would be highly appreciated!

A summary of what's been done follows. Boardgamegeek page for rules etc: https://boardgamegeek.com/boardgame/167791/terraforming-mars

Implemented:
- [x] Base game board, corporations and project cards + corporate era corporations and project cards - using resources from https://ssimeonoff.github.io/ an amazing website for TM!
- [x] Most actions on cards implemented as variations and parameterized options of several action classes
- [x] (hopefully) intuitive GUI for human play
- [x] Winning conditions (though not yet displayed in GUI)
- [x] Solo play (set just 1 player):
      - start with 14 TR + 0 production (Corporate Era)
      - game ends after generation 14 (regardless of global parameters)
      - player wins if Mars is terraformed by the end of the game
      - setup: X cities + 1 adjacent greenery each are placed randomly on the map on legal positions by neutral player (gives X money production if Tharsis Republic corporation is chosen)
      - cards: if a card reduces other player's resources/production, they can always be played as if a 2nd player with infinite resources exists (therefore these can still be played for the benefits, and can ignore the negative effects).
      - not fully play-tested
- [x] Corporate Era 0 starting production
- [ ] Some actions are not yet implemented:
    - [x] Selling projects for 1MC/project
    - [ ] TITANIUM may be used as if...  :  
    - [ ] STEEL MAY BE USED as if..  : 
    - [ ] temporary discounts or effects (next card / this generation only)
    - [ ] but (do something else) if you have ...
- [ ] If the game goes on long enough to use up all the project cards, the deck does not yet reshuffle with the discards.
- [ ] Final Greenery placement

Known issues:
- [x] Bugs in state copies, random player recommended for human testing
- [x] A strange city bug where one player can place infinite cities for free (unkown cause) [Edit: caused by Tharsis Republic corporation, first actions and active actions were not flagged properly as done - now fixed, state saved on the card instead of actions]
- [x] Sometimes resources to pay are not calculated correctly when multiple options are available; reason for this yet unknown, as it only happens in some cases, and this will cause the game to crash as the player will end up in a situation where they can't pay for the action with remaining resources. Similar issue: sometimes paying for a card does not actually play the card, but you lose the resources.
- [x] Actions that would cause specific card choices (e.g. add resource to another card) are technically implemented, but don't work.
- [x] Actions to look at top cards and buy/take them not working
- [x] Action fail: place X>1 tiles (only places the first)
- [x] Effects for money back and discounts on cards sometime not working (corporations ok)
- [x] Bonuses from global parameters don't apply anymore
- [x] Winner and final scoring not diplayed in GUI
- [ ]  Actions menu flickering

image

image

image


rdgain:
Some bug fixes:
- Pay for action working correctly and removes cost requirement for action to be executed after it's paid for.
- Resources removed not a requirement (usually it's 'up to' phrasing) unless production		
- Card resource requirement looking at cards in hand instead of 'card' resources
	issue: Mars University: FAIL need 1 card
	issue: Comet: FAIL need 3 plant (should not be a requirement)
- Event tags counted
	issue: Media Archives calculating wrong number of events played by all players
- Decomposers card parameter fix (add 1 microbe, not 3)
- 'action cost' removed for adding/removing resources on cards
	issue: Nitrite reducing bacteria: action of 3 microbes -> TR not working
- Temperature parameter bonuses displayed in the right spaces
- Parsing of AddResourceOnCard to not hold resources if 'another', not only if 'any'
	issue: Aerobraked Ammonia Asteroid: should not be in 'active cards' or hold resources (add-2-Microbe-another)
- Action menu update only on new action or turn order change
	issue: GUI menu flickering: should only update once action is taken. Send action taken flag to GUI.
- Added tostring to DuplicateImmediateEffect action
	issue: Robotic Workforce: Copy card action: shows card in GUI, but only action available is 'pass'
- Fixed: Decomposers: not triggering with other tags played?
- Fixed: Search for life: not showing tag of card to draw, and last drawn from deck in GUI.

PR #169, opened 2021-08-16
Expert iteration and various MCTS additions

hopshackle:
1) Added ParameterFactory to support Game Parameter initisalisation from json files
 - a game parameter file can now be passed to GameReportII 

2) Added MultiNTBEA in ParameterSearch
 - GameMultiPlayerEvaluator.java; plus see updated --help documentation in ParameterSearch.java, which now defaults to use MulitNTBEA unless an opponent is specified (MultiNTBEA optimises all agents in a game in parallel, potentially avoiding issues with optimising to a trivial, random player)

3) First version of Dice Monastery (v1.8 of rules)

4) Added some standard 'action-based' heuristics:
These rely on having a fixed hash code for each Action as a key to look up a simple Advantage (numerical value of 'good' that action is). Think of it as a MAST-style heuristic.
- ActionAdvantageHeuristic - this is initialised from a file of hashCode -> double values (so, fixed)
- LearnedAdvantage - this can be used to learn from one game to the next when using MCTS, learning from the Advantages calculated from the MCTS Tree after each decision is made. This can then be used in Rollout, to decide Node Expansion Order, or as an Opponent Model in later games.

5) Various MCTS  variants and parameters added:
 - Added 'ExpertIteration' facility to record state information and the associated statistics for each node in the tree - to be used for learning useful stats and policies (currently only used for the ActionAdvantageHeuristic mentioned above...but much space for expansion)
 - Added MAST options to MCTS (as in Bjornnson & Finnson 2008). These Action-specific values are keyed by the Action hash code, and can be learned either from the Tree actions, Rollout actions, or both.
 - Added AdvantageFunction support that enables a class to be specified to split the order of node expansion
 - Added the ability to specify classes for both Opponent Models and Rollout Policies. These must be classes with either a non-arg constructore, and/or a single Stinrg constructor
 - Added Progressive Widening to MCTS parameters - which then uses the specified Advantage Function

6) RoundRobinTournament updated to have the same ability as GameReportII to add in Listeners to report on game progress and statistics


src/main/java/games/dicemonastery/DiceMonasteryConstants.java
```diff
@@ -0,0 +1,200 @@
...
+                    return WINTER;
+                case WINTER:
+                    return SPRING;
+                default:
+                    throw new AssertionError("Should not be processing " + this);
+            }
+        }
+    }
+
+    public final static int[] RETIREMENT_REWARDS = {5, 4, 4, 3, 3, 2, 2, 2};
```
rdgain, line 118: Parameters instead? This and the other rewards around here.

src/main/java/games/dicemonastery/DiceMonasteryTurnOrder.java
```diff
@@ -0,0 +1,321 @@
...
+
+    public DiceMonasteryTurnOrder(int nPlayers, DiceMonasteryParams params) {
+        super(nPlayers, params.YEARS);
+    }
+
+    private DiceMonasteryTurnOrder(int nPlayers) {
+        super(nPlayers);
+    }
+
+    Season season = SPRING;
```
rdgain, line 29: This seems like a game state variable, not turn order?

rdgain (approved):
Looks good! Only a couple minor comments.

src/main/java/games/dicemonastery/DiceMonasteryTurnOrder.java
hopshackle: Interesting point. I have treated TurnOrder as encapsulating all of the state associated with what turn it is (so the GameState does not store this, but just uses the internal TurnOrder object for this), as well as the logic of when to move the turn forard / change the current player.
I can move this out if it breaks a TAG design principle.

hopshackle:
Further update addresses Raluca's review points about parameterisation (partially):
- Extracted AbstractGameData from Pandemic into the core framework (AbstractGameState)
- added createGameInstance() method that will take a GameParams object
- added ability in GameReport and RRTournament to specify a GameParams json file to be used for a game
- Upgraded DiceMonastery to v1.10 of the rules
- Moved much (not all) of the Dice Monastery config into a new DiceMonasteryParams object, and to data/decks.json 
- Added a Deck<T>.add(Collection<T> newCards) helper function
- Added ability to specify bucket types for a learned advantage heuristic (and generalised this to any value heuristic)
- MCTS update to optionally work with strict 'Advantage' instead of Value - this means we need to record the starting value at rollout.

PR #170, opened 2021-09-26
Addition of Battlelore

Kaxell:
Hello! Please let me know if any changes are required. The game is executable through BattleloreGame.java.


src/main/java/core/Game.java
```diff
@@ -556,13 +556,16 @@ public static Game runOne(GameType gameToPlay, List<AbstractPlayer> players, lon
...
+
             if (ac != null) {
                 // Create GUI (null if not implemented; running without visuals)
                 GamePanel gamePanel = new GamePanel();
                 frame.setContentPane(gamePanel);
                 gui = gameToPlay.createGUIManager(gamePanel, game, ac);
             }
-            frame.setFrameProperties();
+
+            frame.setFrameProperties(runWithVisuals);
```
rdgain, line 568: Unnecessary, revert. If running with visuals (ac != null), then gui will not be null and the frame should always be drawn on, so it should just be always visible.

src/main/java/games/battlelore/BattleloreForwardModel.java
```diff
@@ -0,0 +1,245 @@
...
+public class BattleloreForwardModel extends AbstractForwardModel {
+    @Override
+    protected void _setup(AbstractGameState initialState) {
+        Random random = new Random(initialState.getGameParameters().getRandomSeed());
+        BattleloreGameParameters gameParams = (BattleloreGameParameters) initialState.getGameParameters();
+        BattleloreGameState gameState = (BattleloreGameState)initialState;
+        BattleloreData _data = gameState.getData();
+
+        for (int i = 0; i  < gameState.playerCount; i++) {
+            if (gameState.playerCount != 2) {
```
rdgain, line 32: This doesn't need to be in the for loop, and it should raise an exception, not just a quiet error message.

src/main/java/games/battlelore/BattleloreForwardModel.java
```diff
@@ -0,0 +1,245 @@
...
+        BattleloreGameState state = (BattleloreGameState) currentState;
+        BattleloreGameParameters gameParams = (BattleloreGameParameters) currentState.getGameParameters();
+        action.execute(currentState);
+
+        int playerId = state.getCurrentPlayer();
+        Unit.Faction playerFaction = playerId == Unit.Faction.Dakhan_Lords.ordinal() ?
+                Unit.Faction.Dakhan_Lords : Unit.Faction.Uthuk_Yllan;
+
+        switch (state.getGamePhase().toString()) {
+            case "CommandAndOrderStep":
```
rdgain, line 69: This should use the enum, not a string, much easier to mess up the string name and I believe it might be a costlier comparison too.

src/main/java/games/battlelore/BattleloreForwardModel.java
```diff
@@ -0,0 +1,245 @@
...
+                break;
+        }
+
+        if (checkGameEnd((BattleloreGameState) currentState, playerId)) {
+            currentState.setGameStatus(Utils.GameResult.GAME_END);
+            registerWinner(state, playerId);
+            return;
+        }
+
+        int roundExceedThreshold = 100;
```
rdgain, line 94: This should be a game parameter.

src/main/java/games/battlelore/BattleloreGameState.java
```diff
@@ -0,0 +1,284 @@
...
+        LoreStep
+        //Last three steps can be implemented later
+    }
+
+    public enum UnitType {
+        Decoy, BloodHarvester, ViperLegion, CitadelGuard, YeomanArcher;
+    }
+
+    int playerCount;
+    Random random;
```
rdgain, line 35: It doesn't seem like this is used.

src/main/java/games/battlelore/BattleloreGameState.java
```diff
@@ -0,0 +1,284 @@
...
+    BattleloreGameParameters parameters;
```
rdgain, line 36: There is already a parameter variable in the super class, should just use that.

src/main/java/games/battlelore/BattleloreGameState.java
```diff
@@ -0,0 +1,284 @@
...
+
+    int numberOfRounds;
```
rdgain, line 38: Rounds and turns should be handled in a TurnOrder object, not in the game state.

src/main/java/games/battlelore/BattleloreGameState.java
```diff
@@ -0,0 +1,284 @@
...
+                        counter++;
+                    }
+                }
+            }
+        }
+        return possibleLocations;
+    }
+
+    public int[][] GetPossibleTargetUnits(MapTile attackUnit) {
+        int[][] possibleLocations = new int[gameBoard.getWidth()][2];
```
rdgain, line 186: A hard-coded '2' here that should be either nPlayers, or a game parameter

src/main/java/games/battlelore/actions/AttackUnitsAction.java
```diff
@@ -0,0 +1,120 @@
...
+import games.battlelore.components.Unit;
+
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Random;
+
+public class AttackUnitsAction extends AbstractAction {
+    private Unit.Faction playerFaction;
+    private MapTile attacker;
```
rdgain, line 18: This is a component, and functionality will break if passed by reference. Should use component ID instead, and then find it in the game state when executing the action (findComponentByID method on the base game state class)

src/main/java/games/battlelore/actions/AttackUnitsAction.java
```diff
@@ -0,0 +1,120 @@
...
+        else {
+            ArrayList<Unit> attackerUnits = state.getBoard().getElement(attacker.getLocationX(), attacker.getLocationY()).GetUnits();
+            ArrayList<Unit> defenderUnits = state.getBoard().getElement(defender.getLocationX(), defender.getLocationY()).GetUnits();
+
+            //COMBAT SEQUENCE: Roll a dice
+
+            int defeatedEnemyCount = 0;
+            CombatDice dice = new CombatDice();
+
+            for (int i = 0; i < 3; i++) {
```
rdgain, line 46: '3' should be a game parameter

src/main/java/games/battlelore/actions/MoveUnitsAction.java
```diff
@@ -0,0 +1,92 @@
...
+
+
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.Objects;
+
+public class MoveUnitsAction extends AbstractAction
+{
+    private Unit.Faction playerFaction;
+    private MapTile tile;
```
rdgain, line 25: Similar, this should be a component ID and not the reference to the component

src/main/java/games/battlelore/actions/SkipTurnAction.java
```diff
@@ -0,0 +1,93 @@
...
+import games.battlelore.components.MapTile;
+import games.battlelore.components.Unit;
+import games.dotsboxes.AddGridCellEdge;
+
+import java.util.Objects;
+
+public class SkipTurnAction extends AbstractAction {
+
+    private Unit.Faction playerFaction;
+    private MapTile units;
```
rdgain, line 16: Similar, this should be a component ID

src/main/java/games/battlelore/components/MapTile.java
```diff
@@ -0,0 +1,168 @@
...
+
+    public int getLocationX() {
+        return locationX;
+    }
+
+    public int getLocationY() {
+        return locationY;
+    }
+
+    public MapTile(int locationX, int locationY, ArrayList<Unit> units, Terrain terrain, int componentID) {
```
rdgain, line 25: This type of constructor (with component ID) should not be used for things other than making copies. Component IDs are set automatically for all components in the game, so that each component has a unique ID; should not expose this externally.

src/main/java/games/battlelore/components/Terrain.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class Terrain extends Component {
+
+    public Terrain(Utils.ComponentType type, String name) {
+        super(type, name);
+    }
+
+    @Override
+    public Component copy() {
+        return null;
```
rdgain, line 14: You likely don't need this class at all, since it doesn't contain anything that Component doesn't - either way, returning null for copies is not indicated.

src/main/java/gui/GUI.java
```diff
@@ -54,13 +54,13 @@ public GUI(ActionController ac, int maxActionSpace) {
...
 
-    public void setFrameProperties() {
+    public void setFrameProperties(boolean runWithVisuals) {
         // Frame properties
         this.wi = new WindowInput();
         addWindowListener(wi);
         revalidate();
         pack();
-        this.setVisible(true);
+        this.setVisible(runWithVisuals);
```
rdgain, line 63: Not necessary, revert

Kaxell:
Hey @rdgain , I've done the review fixes, and it is ready for confirmation. Please let me know if anything else is required.

src/main/java/core/Game.java
Kaxell: done in part 1

src/main/java/games/battlelore/BattleloreForwardModel.java
Kaxell: done in part 1

src/main/java/games/battlelore/BattleloreForwardModel.java
Kaxell: done in part 1

src/main/java/games/battlelore/BattleloreForwardModel.java
Kaxell: done in part 5

src/main/java/games/battlelore/BattleloreGameState.java
```diff
@@ -0,0 +1,264 @@
...
+
+    public enum UnitType {
+        Decoy, BloodHarvester, ViperLegion, CitadelGuard, YeomanArcher;
+    }
+
+    int numberOfRounds;
+    int[] playerScores;
+    GridBoard<MapTile> gameBoard;
+    List<Unit> unitTypes;
+    BattleloreGameParameters parameters;
```
rdgain, line 39: Re-adding comment: There is already a parameter variable in the super class, should just use that.

src/main/java/games/battlelore/actions/AttackUnitsAction.java
rdgain: Clarifying this: the class itself (AttackUnitsAction) needs to store primitive types (e.g. attack unit tile ID, rather than the tile itself) - the actual objects can then be retrieved by ID when executing the action.

src/main/java/games/battlelore/cards/CommandCard.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private int orderCount;
+    private int bonusAttackCount;
+    private int bonusLorePoints;
+    private int bonusMoveCount;
+    private Boolean hasCondition;
+    protected String title;
+    protected String explanation;
+
+    public CommandCard(String title, String explanation, int orderCount, int bonusAttackCount, int bonusLorePoints, int bonusMoveCount) {
```
rdgain, line 49: The constructors here should all call a super() constructor since you are extending another class.

src/main/java/games/battlelore/BattleloreForwardModel.java
```diff
@@ -0,0 +1,242 @@
...
+    protected AbstractForwardModel _copy() {
+        return new BattleloreForwardModel();
+    }
+
+    /**
+     * Checks if the game ended.
+     * @param gameState - game state to check game end.
+     */
+    private boolean checkGameEnd(BattleloreGameState gameState, int playerId) {
+        int WIN_SCORE = 4;
```
rdgain, line 224: A game parameter

rdgain (changes requested):
I resolved previous comments addressed, only a couple more issues.

src/main/java/games/battlelore/BattleloreGameState.java
rdgain: To clarify this as well: the TurnOrder object already keeps track of turns and rounds, so this is duplicating functionality that could be accessed from the TurnOrder (and/or adjusted as needed in the custom TurnOrder object, which at the moment adds nothing to the base class)

Kaxell:
done 💯

rdgain (approved):

PR #172, opened 2021-12-01
MultiTree MCTS, UCB-Tuned and Nested tuning of rollout policies and heuristics

hopshackle:
There are two main pieces of new functionality here:
A) New MCTS options
B) Improved ability to auto-tune player parameters

### A) New MCTS options

1) MultiTree MCTS
Two new options for `opponentTreePolicy` have been added:
`MultiTree` - This constructs one tree per player, with the transition in each tree only being the determined by the actions of the relevant player. This is inspired by Multiple Observer MCTS (Cowling et al. 2012), but with an even stronger constraint that each opponent can only 'see' their own moves, and hence can only condition future actions on these. The idea is that we have a stronger opponent model than random, while retaining a much deeper effective search than having all opponent actions in a single tree (as with `MaxN` and `Paranoid`).
`MultiTreeParanoid` - this is exactly as `MultiTree` with the addition of the paranoid assumption; the reward function for all other players is assumed to be minimising our score ('us' being the player currently using MCTS to decide their next move).  
    In terms of implementation, on each iteration we may add up to one node in each of the player trees. As actions are taken we may be in the rollout phase for some players, while others are still in the tree.

2) UCB1-Tuned
A new option in `treePolicy` of `UCB_Tuned` implements UCB1-Tuned from Auer et al. 2002. This reduces the exploration term in UCB1 based on the (lack of) variance in the observed rewards to date. This may lead to faster focus on good options in less noisy environments.

3) Refactor of `openLoop` and `redeterminise` flags
These have been amalgamated into a new MCTS parameter of `information` that takes values of `Closed_Loop`, `Open_Loop` and `Information_Set`. This is clearer and removes the nonsensical setting of `OL=false, R=true`, as redeterminising with a Closed Loop approach is meaningless.

4) Added new MCTS parameter for `maintainMasterState` that defaults to `false`. When `true` this maintains a pristine copy of the game state when the node was created during expansion. This is essential for some features, such as Expert Iteration data gathering, but can be a big drain on resources as we have to do an expensive deep copy on each such expansion. (This used to be the default, now switched off by default.)

5) Added two new MCTS parameters for minor universal tweaks:
`normaliseRewards` - defaults to `true`. This will maintain a record of low/high rewards detected during iterations to date, and then scale all rewards to be in [0.0, 1.0] using these. This means that ideally MCTS functionality is now invariant to the scale of the game scoring system. (And hence much less need to tune C.)
`nodesStoreScoreDelta` - default to `true`. The default will now be to take the value of a rollout as the *difference* in score/heuristic that results compared to the starting state (instead of the absolute value).

6) Added explicit parameters in MCTS to define each of:
`rolloutPolicyParams` - the details the TunableParameters that can be used to instantiate an `AbstractPlayer` to act as rollout policy (the main one implemented is `OSLAHeuristic`, which uses a configurable heuristic in OSLA).
`oppModelParams` - as above, but defines the `AbstractPlayer` used by all opponents. It is not currently possible to define a different model for each opponent separately.
`heuristic` - the heuristic function to be used to value the state at the end of each rollout for back-propagation
`opponentHeuristic' - the heurisic function used by opponents (another form of opponent modelling being to change this).

### B) Improved Tuning options

7) It is now possible (as per point 6) above) to tune a rollout policy or opponent model via NTBEA. (Previously it was only possible to tune the heuristic function).

8) Flexibility of parameter optimisation is now enhanced to allow full nesting in JSON. 
For example it is possible to tune the parameters using in our heuristic function, as well as the parameters of some opponent model via an OSLA agent for example, as well as key MCTS parameters. (Previously this actually failed if you tried more than one level of recursive nesting.)

9) Not everything in a JSON parameters file now needs to subclass `TunableParameters`. If a thing is instantiated from JSON that is not such a sub-class, then we stop further recursion to find tunable parameters, but still instantiate the thing. The bets current example is `heuristic` or `opponentHeuristic` in MCTS. These can provide a specific class that implements `IStateHeuristic` without also needing to themselves be `TunableParameters`. (See next point for where this can be useful.)

10) Two standard heuristic functions provided for `PureScoreHeuristic` and `ScoreHeuristic`. The former ignores win/lose conditions and just tries to maximise the inherent game score. The latter treats a win as a 50% bonus to score, with a 50% penalty for a loss. These are useful to plug in as heuristics as MCTS if the default game heuristic is rather questionable (or for a useful baseline comparison).

### C) Other bits and pieces

11) Add a standard optional function of `getTiebreak(int player)` to `AbstractGameState`. This is used in `getOrdinalPosition(int player)` to break ties not represented in the formal score. This is used if we are optimising parameters using an `Ordinal` condition that treats 1st as better than 2nd as better than 3rd (and so on).

12) Added a default heuristic evaluation to SushiGo that simply used the game score (previsouly this always returned 0.0, which led to random MCTS behaviour). 

13) Refactored some Poker, BattleLore and LoveLetter actions to avoid unnecessary copying of immutable objects, and to correct a couple of incorrect hash() implementations.


resources/manifests/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,3 @@
+Manifest-Version: 1.0
+Main-Class: evaluation.GameReport
```
rdgain, line 2: Let's keep this out of repo to avoid build conflicts.

src/main/java/core/AbstractPlayer.java
```diff
@@ -32,6 +32,9 @@ public final int getPlayerID() {
     public final AbstractForwardModel getForwardModel() {
         return forwardModel;
     }
+    public void setForwardModel(AbstractForwardModel model) {
```
rdgain, line 35: I don't like the idea of this, having it public would break the confidence that the FM received is the proper one for the game played, as set in the Game class. I could see some uses for it being public, is there something specific in this PR that needs it?

src/main/java/core/interfaces/ITunableParameters.java
```diff
@@ -140,6 +140,6 @@ default void setParameterValues(Map<?, Object> values) {
      * @param json      The raw JSON detailing the parameters
      * @return          The instantiated object
      */
-    ITunableParameters registerChild(String name, JSONObject json);
+    Object registerChild(String name, JSONObject json);
```
rdgain, line 143: Why is this an Object now? Seems very general and I'm afraid of things breaking it.

src/main/java/evaluation/ParameterSearch.java
```diff
@@ -251,7 +252,7 @@ public static void runMultiNTBEA(NTupleSystem landscapeModel, String[] args) {
         Pair<Pair<Double, Double>, double[]> bestResult = new Pair<>(new Pair<>(Double.NEGATIVE_INFINITY, Double.NEGATIVE_INFINITY), new double[0]);
         for (int mainLoop = 0; mainLoop < repeats; mainLoop++) {
             landscapeModel.reset();
-            Pair<Double, Double> r = runNTBEA(null, evaluator, searchFramework, iterationsPerRun, iterationsPerRun, evalGames, verbose);
+            Pair<Double, Double> r = runNTBEA(null, evaluator, searchFramework, iterationsPerRun, 500, evalGames, verbose);
```
rdgain, line 255: This seems like a backwards change of variable to hard-coded number, any reason for it?

src/main/java/games/battlelore/BattleloreForwardModel.java
```diff
@@ -1,31 +1,30 @@
...
 import utilities.Utils;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 public class BattleloreForwardModel extends AbstractForwardModel {
+
+    int maxTurnsToPlay = 100;
```
rdgain, line 22: This looks like it should be a parameter - but also unrelated to this PR so can be done separate, just wanted to add a quick note.

src/main/java/games/loveletter/LoveLetterForwardModel.java
```diff
@@ -246,6 +248,9 @@ public boolean checkEndOfGame(LoveLetterGameState llgs) {
             int bestValue = 0;
             for (int i = 0; i < llgs.getNPlayers(); i++) {
                 if (llgs.getPlayerResults()[i] != Utils.GameResult.LOSE) {
+                    if (llgs.playerHandCards.get(i).peek() == null) {
```
rdgain, line 251: What's the purpose of this?

src/main/java/games/loveletter/LoveLetterHeuristic.java
```diff
@@ -32,11 +32,12 @@ public double evaluateState(AbstractGameState gs, int playerId) {
         LoveLetterParameters llp = (LoveLetterParameters) gs.getGameParameters();
         Utils.GameResult playerResult = gs.getPlayerResults()[playerId];
 
-        if (playerResult == Utils.GameResult.LOSE)
-            return -1;
-        if (playerResult == Utils.GameResult.WIN)
-            return 1;
-
+        if (!gs.isNotTerminal()) {
```
rdgain, line 35: Is this necessary? If a player's result was set to lose/win, then they're no longer playing so don't need to do the more complex calculations

rdgain (changes requested):
Looks good, some small questions.

resources/manifests/META-INF/MANIFEST.MF
hopshackle: Completely! Careless of me to let that sneak in.

src/main/java/core/AbstractPlayer.java
hopshackle: The specific reason in this PR is in MCTSPlayer when it contains distinct rollout strategies or opponent models that  themselves are implementations of AbstractPlayer (so can be much more sophisticated than random). These therefore also need to be initialised with the forward model by Game...but are nested within MCTSPlayer, so the setForwardModel() method here is recursive, with MCTSPlayer forwarding on the model to its internal components when it is received from Game.

One option I toyed with was including this within initializePlayer(), to be passed at the same time as the initial game state; but that is also public so not an improvement.

Another option might be to have Game pass ForwardModel at the same time as AbstractGameState when it calls getAction()? (although again, this is a public method...)

(Also - I think it might be quite reasonable for an AbstractPlayer to use a non-perfect forward model.)

src/main/java/core/interfaces/ITunableParameters.java
hopshackle: The javadoc is a bit out of date, so I'll correct this. (and it's taken me a while to remember why I did this - so definitely needs documenting)

The underlying reason is that registerChild() is used whenever we are extracting settings for an ITunableParameters from JSON. It is called when we find a JSONObject as a property in the JSON file. This change is for point 9) in the changes, specifically:

"Not everything in a JSON parameters file now needs to subclass TunableParameters. If a thing is instantiated from JSON that is not such a sub-class, then we stop further recursion to find tunable parameters, but still instantiate the thing. The best current example is heuristic or opponentHeuristic in MCTS. These can provide a specific class that implements IStateHeuristic without also needing to themselves be TunableParameters. (See next point for where this can be useful.)"

src/main/java/evaluation/ParameterSearch.java
hopshackle: Oops - left in from debugging. Fixed.

src/main/java/games/battlelore/BattleloreForwardModel.java
hopshackle: Easy to fix at the same time. Done.

src/main/java/games/loveletter/LoveLetterForwardModel.java
hopshackle: Ahem...left in from when I was debugging a problem in LoveLetter with MultiTree MCTS. Forgot to remove it...now done.

src/main/java/games/loveletter/LoveLetterHeuristic.java
hopshackle: Odd - the change log gives a strange view. If you look at the actual full code you'll see this is exactly the case.

src/main/java/core/AbstractPlayer.java
rdgain: I'm more worried of a case where a player is accidentally given a FM for a different game through this, which would make all sorts of weird things happen, some not obvious - and if we open this in competition format, it'd also raise the question of can any player hack another player's FM and break the confidence we'd have in fairness there? But I can see potential for the other use cases you've mentioned, maybe add a warning in docs / an issue for us to check in future competition setting?

hopshackle: I'm much less worried about this. This fake forward model 'attack' would require the other player to get a reference to our AbstractPlayer, and I don't see how this is possible, as this is held within the Game, and the at no point is a reference to the Game passed to the players - they get the AbstractGameState only, which also does not have any reference to the Game.
...am I missing something?

rdgain: Right, at the moment this is not the case, but with the framework expanding and more things becoming public - and there being more and more things, easy to start to lose track of details -, I can see this becoming an issue. Adding a note to the docs would give piece of mind.

hopshackle: Ok - added some notes on this.

For competitions in the future we could perhaps pass the ForwardModel to the AbstractPlayer along with the AbstractGameState in getAction(). That way any 'corrupted' agent could always refresh its copy.

rdgain: True, that's a good option! Thanks! :) Approved for merge!

rdgain (approved):

PR #173, opened 2022-01-03
Game tuning

hopshackle:
Implements all the required stuff to use NTBEA for tuning the parameters of a game to meet some defined objective function.

New documentation now online at http://www.tabletopgames.ai/wiki/agents/GameTuning#ParameterSearch


src/main/java/games/tictactoe/TicTacToeGameState.java
```diff
@@ -57,7 +60,11 @@ protected double _getHeuristicScore(int playerId) {
      */
     @Override
     public double getGameScore(int playerId) {
-        return 0;
+        if (playerResults[playerId] == Utils.GameResult.WIN)
```
rdgain, line 63: In ExplodingKittens update it's 1 for WIN and -1 for LOSE, this should be somewhat consistent. In fact, the GameResult enum has values associated, I'd suggest using those instead of arbitrary numbers different across games.

src/main/java/utilities/Utils.java
```diff
@@ -179,8 +153,8 @@ public static void swap(int[] input, int a, int b) {
     public static double normalise(double a_value, double a_min, double a_max) {
         if (a_min < a_max)
             return (a_value - a_min) / (a_max - a_min);
-        else    // if bounds are invalid, then return same value
-            return a_value;
+        else    // if bounds are invalid, then return 0.0
+            return 0.0;
```
rdgain, line 157: I'm not sure if this makes sense. If you're trying to normalise 5 between [7,10], should you get the same result as if you were trying to normalise 20 in the same interval? Probably best actually to raise an error...

rdgain (commented):
Thanks for all the work on this! 

I'm a bit confused about the general setup. Do I understand that this is set up as follows:
- Needs a game, tunable parameters and a search space defined.
- Also needs a players defined (as opponents)... and no central player, right?
- Runs the game once for each setting explored by NTBEA, with players given (btw: some leftover functions point to agent only, e.g. getAgent function on the ITPSearchSpace)
- Evaluates with heuristic which receives the game object, which is the fitness for whatever parameter configuration tested

Is there an easy way to give more control over the evaluation (not necessarily as a replacement, but as a more customisable option). E.g.:
- Run multiple games per parameter configuration, extracting custom statistics for each run
- Human player included in evaluation as player / judge (so, option to visualise games or stats)

Would this generally mean just creating custom GameEvaluator-like classes?

src/main/java/core/AbstractPlayer.java
```diff
@@ -33,6 +33,17 @@ public final AbstractForwardModel getForwardModel() {
...
+     * This is used by Game, and also when an AbstractPlayer is a component of another agent
+     * (for example in MCTSPlayer, where the heuristic rollout policy could itself be an AbstractPlayer
+     * that needs a copy of the forward model)
+     *
+     * Anything that has a reference to an AbstractPlayer can therefore pass a ForwardModel.
+     * In competition mode this it is not currently possible for opponents to get a link to your
+     * object...and this should be avoided.
+     * @param model
+     */
     public void setForwardModel(AbstractForwardModel model) {
```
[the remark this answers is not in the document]
hopshackle: I've done this for several - but not for setForwardModel(), as I only added it *because* I need to then override it in MCTSPlayer :-)

src/main/java/core/AbstractPlayer.java
```diff
@@ -83,4 +94,9 @@ public IStatisticLogger getStatsLogger() {
     public void setStatsLogger(IStatisticLogger logger) {
         this.statsLogger = logger;
     }
+
+    // Override if a Player maintains any state at all
+    public AbstractPlayer copy() {
```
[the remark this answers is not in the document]
hopshackle: Done.

src/main/java/games/tictactoe/TicTacToeGameState.java
hopshackle: Cool! I have taken that as licence to re-write the main sets of these values. (Having LOSE, GAME_ONGOING and DRAW all equal to 0.0 annoyed me.) Now standardised across the board!

src/main/java/utilities/Utils.java
hopshackle: Agreed. Done.

rdgain (approved):

PR #174, opened 2022-01-19
Random seed

hopshackle:
The main change in this is to ensure that when running GameReport and RoundRobinTournaments that the game seed is set to a new value for each game.

Previously the same random game seed was used for every game...which meant that every game used the same startnig hands and so on. (Gave some very funny results in Diamant in particular)

Also tidies up some duplicate code in Game


rdgain (changes requested):
I prefer the main method in the Game class to be at the bottom so it can be easily found and adjusted when running. Otherwise good!

rdgain (approved):

PR #176, opened 2022-02-01
Generalised parameter search

rdgain:
- Parameter search generalised to any EvoAlg, not only NTBEA
- Adds an enumeration for search frameworks in ParameterSearch class, and a method to construct these appropriately. 
- Skeleton for MapElites included, with some methods left open on TODOs. Can specify arbitrary number of behaviours, where each has a name, and a list of possible values (if comparable, the closest one is used for indexing; if not, equality is checked to used for indexing in the map). MapElites implements the EvoAlg interface, with some adjustments.
- A new method added in GameEvaluator which returns not only a value, but also a map from behaviour to value. These behaviour names and value types should match to those defined in ParameterSearch (though no explicit checks for this are included yet).

Better full integration of other optimisers could be expanded in the future (this still assumes model-based approaches and makes some other assumptions MapElites ignores, NTBEA has quite a rigid framework with some odd decisions behind it too to base this on).

### DIFF
--- a/src/main/java/evaluation/GameEvaluator.java
+++ b/src/main/java/evaluation/GameEvaluator.java
@@ -6,6 +6,7 @@ import core.interfaces.IStateHeuristic;
 import core.interfaces.IStatisticLogger;
 import evodef.*;
 import games.GameType;
+import utilities.Pair;
 import utilities.SummaryLogger;
 
 import java.util.*;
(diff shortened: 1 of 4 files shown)


src/main/java/evaluation/GameEvaluator.java
```diff
@@ -147,6 +148,59 @@ public double evaluate(int[] settings) {
...
 
+    /**
+     * Method used to return more information about the game run (e.g. for map elites)
+     *
+     * The behaviours specified at the end of this method should match (in name, and type of data recorded) to those defined in:
+     * @see ParameterSearchMapElites#getSearchFramework
+     */
+    public Pair<Double, TreeMap<String, Object>> evaluateWithStats(int[] point) {
+        Object configuredThing = searchSpace.getAgent(point);
+        boolean tuningPlayer = configuredThing instanceof AbstractPlayer;
```
hopshackle, line 159: I'm getting conniptions at the oodles of exactly duplicated code from evaluate()...any chance that this could extracted out to a helper function to reduce duplication?

hopshackle (changes requested):

PR #179, opened 2022-02-07
Game tracking

hopshackle:
Main changes:

1) Frontend is now reworked to have the GUI running in a separate thread to the Game, and with the Game not being aware whether it is running in a GUI or not. This This enables:

2) Various button pairs created/merged:
Play/Stop
Pause/Resume
Show Self / Show All
A new 'Next Action' button then allows the user (when in Pause mode) to step through the AI actions one by one instead of seeing them go by in a blur.

3) When an AI is moving, Frontend will make 20 (parameter) calls to the agent to get a distribution over the actions it might take, and displays this on the GUI (if in pause mode), as green tinges on the action buttons. This does not work for games which do not use the core code for these buttons - i.e/. TicTacToe, Dots and Boxes and Love Letter.

4) A human user may now choose to override the action of an AI by clicking on an action when it is their turn (as long as you are in Pause Mode and have Show All enabled).

5) Game updated to retain a main() function to allow a game to be run with or without GUI via that old route. This does not have all the extra functionality above.

5) Diamant Game state now amended to record a playerActionRecord for when each player left the cave on each turn. For use in game evaluation functions.


src/main/java/core/Game.java
```diff
@@ -284,6 +292,62 @@ public static void recordPlayerResults(TAGStatSummary[] statSummaries, Game game
...
+     * some games for some screen sizes due to the vagaries of Java Swing...
+     * <p>
+     * Test class used to run a specific game. The user must specify:
+     * 1. Action controller for GUI interactions / null for no visuals
+     * 2. Random seed for the game
+     * 3. Players for the game
+     * 4. Mode of running
+     * and then run this class.
+     */
+    public static void main(String[] args) {
```
rdgain, line 306: Mm I still don't like the main method being in the middle of this file.

src/main/java/core/CoreConstants.java
```diff
@@ -25,7 +25,7 @@
     public final static int backgroundImgHash = Hash.GetInstance().hash("backgroundImg");
 
     public enum GameEvents {
-        ABOUT_TO_START, GAME_OVER, ROUND_OVER, TURN_OVER, ACTION_CHOSEN, GAME_EVENT
+        ABOUT_TO_START, GAME_OVER, ROUND_OVER, TURN_OVER, ACTION_CHOSEN, ACTION_TAKEN, GAME_EVENT
```
rdgain, line 28: What's the difference between action chosen, and action taken?

rdgain (commented):
Some other questions:
- What's the functionality of the 'show all' button? I probably missed wherever this is explained. :)
- For games where not all actions appear in the action panel (or maybe for all for different ways of communicating the information / accessibility): how hard would it be to add another small panel in the GUI which shows a ranking of all possible actions, or maybe just a top..3 ? The information for this should be already passed to the right methods, it should be just using it to write it up in text format rather than colouring the actions, I think?

src/main/java/core/Game.java
hopshackle: Which is why I deliberately moved it to the end as the last commit on the Pull Request!

src/main/java/core/CoreConstants.java
hopshackle: I added doc on this distintion to the point of their generation, but not here - I'll add it.

ACTION_CHOSEN = before action implemented (so state records information that affected the decision)
ACTION_TAKEN = after action implemented (which is what we want in order to update the GUI, say)

src/main/java/core/Game.java
hopshackle: Added tool tips to the buttons to clarify. Show All shows the actions for all players when it is their turn (so that you can override their decision), while Show Self is the old default that only shows the actions available to the human player (and hence avoids information leakage),

Adding the extra information is not difficult in theory - but I'd rather not do it now as fiddling with the GUI layout tends to take longer than anticipated.

PR #184, opened 2022-04-20
RHEA

Nirase:


src/main/java/players/rhea/RHEAIndividual.java
```diff
@@ -0,0 +1,177 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (!(o instanceof Individual)) return false;
+
+        RHEAIndividual a = this;
+        RHEAIndividual b = (RHEAIndividual)o;
+
+        for (int i = 0; i < actions.length; i++) {
+            if (!a.actions[i].equals(b.actions[i])) return false;
+        }
```
hopshackle, line 166: Hence if two individuals have the same sets of actions, but these led to different states (in a stochastic game, or if we redeterminesed for each), then they will be treated as identical.
Probably correct - checking this is what you intended?

src/main/java/players/rhea/RHEAIndividual.java
```diff
@@ -0,0 +1,177 @@
...
+                    // Individual length increased
+                    length++;
+
+                    // Add value of state, discounted
+                    double score;
+                    if (this.heuristic != null){
+                        score = heuristic.evaluateState(gameStates[i+1], playerID);
+                    } else {
+                        score = gameStates[i+1].getHeuristicScore(playerID);
+                    }
```
hopshackle, line 134: As previous point. getGameScore() may be a better general default

src/main/java/players/rhea/RHEAIndividual.java
```diff
@@ -0,0 +1,177 @@
...
+                    delta += Math.pow(discountFactor, i) * score;
```
hopshackle, line 135: Is this correct - delta appears to add up the discounted value of all scores. 
So consider the case in which we have scores of 1, 2, 2, 4, 3 on successive actions should we not discount the differences (i.e. 1, 0, 2, -1) and add these up to get delta to avoid multiple-counting?

The score is the total current value, not the incremental gain in reward.

src/main/java/players/rhea/RHEAPlayer.java
```diff
@@ -0,0 +1,333 @@
...
+    }
+    /**
+     * Run evolutionary process for one generation
+     * @param stateObs - current game state
+     */
+    private void runIteration(AbstractGameState stateObs) {
+        ElapsedCpuTimer elapsedTimerIteration = new ElapsedCpuTimer();
+
+        //selection
+        population.sort(Comparator.naturalOrder());
```
hopshackle, line 236: Nice touch on comparability!

src/main/java/players/rhea/RHEAPlayer.java
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    @Override
+    public RHEAPlayer copy() {
+        RHEAParams newParams = (RHEAParams) params.copy();
+        newParams.setRandomSeed(randomGenerator.nextInt());
+        return new RHEAPlayer(newParams);
+    }
+
+    private RHEAIndividual crossover(RHEAIndividual p1, RHEAIndividual p2)
```
hopshackle, line 118: A general point on all the crossover methods. What happens if the action sequences are simply incompatible, and the action immediately after the crossover point is illegal?
For a simpe example, consider TicTacToe where we crossover two parents - the first one starts with playing in the middle square...while the second one starts in the top right corner, and plays in the middle second. This could lead to a genome that plays in the middle twice (which is illegal).

src/main/java/players/rhea/RHEAPlayer.java
```diff
@@ -0,0 +1,333 @@
...
+        }
+
+        //crossover & mutation
+        for(int i = 0; i < params.childCount; ++i)
+        {
+            RHEAIndividual[] parents = selectParents();
+
+            RHEAIndividual child = crossover(parents[0], parents[1]);
+            statesUpdated += child.mutate(getForwardModel(), getPlayerID());
+            //fmCalls += child.rollout(child.gameStates[0], getForwardModel(), 0, child.actions.length, getPlayerID());
```
hopshackle, line 253: My understanding of this is that we crossover the genomes (which consist of the actions), and then mutate will pick a random action index, and rollout just from that point forward - starting from the game state stored against that action. At no point (that I can see at the moment) do we rollout using all the actions in the genome, but instead assume that we can just pick up at the stored game state.

I think this may be missing a few things (or possibly I am). Consider the TicTacToe example (from comment in crossover). If we crossover two parent genomes so that we now have something like:
middle - middle - top right - ...

The state stored against the first middle is from parent 1 - and is the start state, so fine.
The state stored against the second middle is from parent 2 - and will have a cross in the middle (and a nought elsewhere. 
The score given to this sequence of actions will just depend (correct?) on the rollout from the state picked at the mutation index - which may not actually include the very first action.

hopshackle (commented):
Nice and clear code!

I've left a few comments that it will probably be useful to talk through early next week if you have the time - my main concern is about what the crossover() and mutate() methods do to detect illegal/incompatible actions.

PR #189, opened 2022-05-27
Added a cost to all neighbours of a BoardNode.

GAIGResearch:
By default, cost is 1.0.
Can access costs by new getNeighbourCost() function.

### DIFF
--- a/src/main/java/core/components/BoardNode.java
+++ b/src/main/java/core/components/BoardNode.java
@@ -10,7 +10,7 @@ import java.util.HashSet;
 
 public class BoardNode extends Component {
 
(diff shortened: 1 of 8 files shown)


src/main/java/core/components/BoardNode.java
```diff
@@ -10,14 +10,16 @@
 
 public class BoardNode extends Component {
 
-    private HashSet<Integer> neighbours;  // Neighbours of this board node, component IDs
+    final double defaultCost = 1.0;
```
rdgain, line 13: Maybe a core parameter if possible to get access to that, or at least static

rdgain (approved):

PR #197, opened 2022-05-31
Descent movement

markotot:
Added 2 types of locations to available actions:
1) Adjacent locations 
2) Points of interests (tokens)

Changes:
- Added remainingMovementPoints to Figure.cs
- Added tile coordinates to GUI display for easier debugging to DescentGridBoardView.cs
- Implemented detection of adjacent squares and detection of points of interest in DescentForwardModel. TODO: use the pathfinding to determine the distance between player and point of interest.

### DIFF
--- a/src/main/java/games/descent2e/DescentForwardModel.java
+++ b/src/main/java/games/descent2e/DescentForwardModel.java
@@ -385,11 +385,13 @@ public class DescentForwardModel extends AbstractForwardModel {
                 }
 
 
-                //PropertyInt figureOnLocation = (PropertyInt)figureNode.getProperty(playersHash);
-                //Figure neighbourFigure = (Figure)dgs.getComponentById(figureOnLocation.value);
-                /*if (figureType.equals(neighbourFigure.getTokenType())){
-                    isFriendly = true;
-                }*/
+                PropertyInt figureOnLocation = (PropertyInt)neighbour.getProperty(playersHash);
+                if (figureOnLocation.value != -1) {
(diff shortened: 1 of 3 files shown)


src/main/java/games/descent2e/DescentForwardModel.java
```diff
@@ -323,35 +323,132 @@ protected List<AbstractAction> _computeAvailableActions(AbstractGameState gameSt
```
rdgain, line 323: Some potential points on things that exist that could simplify some of this logic 
- Utils.getNeighbourhood(x,y,width,height,8-way connection) is a function that returns the list of Vector2D coordinates for adjacent squares (although checking the board node neighbours makes more sense here, just noting this in case it becomes useful at some point)
- Also, all board nodes on the grid have a property that stores the component ID of the figure at that location (-1 if empty I believe, or property is null or something, worth checking FM setup) - at least in the current setup. So can get e.g. a Figure object that occupies a board node by (Figure)gs.getComponentById(((PropertyInt)boardNode.getProperty(playersHash)).value). I think there are some checks for occupancy in your code, this might be an easier way to do that (can then potentially check if figure instanceof Hero/Monster to see if friendly). This might change in the future, but I think nodes should store figure occupancy anyway so some form of this should exist.

markotot: Changed the logic of the code to use gs.getComponentById(((PropertyInt)boardNode.getProperty(playersHash)).value instead

PR #205, opened 2022-06-09
Descent GUI updates (+ other adjustments and bugfixes)

rdgain:
GUI updated to allow human interaction for easier testing.

- GUI shows hero areas as well as the board, with initial stats (drawn on hero card), current attribute values (move points, XP etc) and owned tokens
- Highlighting movement action in GUI shows squares traveled through on the board
- Pretty graphics, fonts, colours
- Flag in DescentGUI for pretty graphics on/off
- Turn order bug fixes
- All base heroes added as JSON options
- Thief and berserker decks for scout/warrior, all 4 archetypes have 1 class available

Some adjustments/bugfixes related to:
- Movement (whole path given to move action to apply travel effects for all spaces and draw in GUI; terrains hold data on attribute value modifiers, e.g. move points spend / damage applied)
- Added "get move points" action
- Acolyte/search action
- Actions that cost action points to execute count this correctly (except for attacking, not touched that)

It also seems things don't work properly if duplicate of same hero is used, so that line commented back in to avoid it. 

image
image


rdgain:
image

PR #216, opened 2022-09-29
Cant stop

hopshackle:
Implementation of Can't Stop

All working including for Human GUI or Console play


GAIGResearch:
Looks great! I played it and defeated Random very decisively.

PR #223, opened 2022-10-18
Fixes from Game AI lab

hopshackle:
1) Added error message if heuristic returns NaN in RHEA/RMHC/MCTS/BasicMCTS - previously this was being stored in the expanded node, and causing a crash later during selection. This now fails early to enable rapid pinpointing of the error.

2) Reversed out the changes for Diamant (Components are back)

3) Corrected RMHC rollout to match RHEA - we discount the *changes* in the score at each time-step, not the full state evaluations, which is tremendously wrong.

4) Fixed CantStop._getHeuristicScore() to obey the contract that it is bounded in [-1, +1]

### DIFF
--- a/src/main/java/games/cantstop/CantStopGameState.java
+++ b/src/main/java/games/cantstop/CantStopGameState.java
@@ -98,7 +98,7 @@ public class CantStopGameState extends AbstractGameState implements IPrintable {
 
     @Override
     protected double _getHeuristicScore(int playerId) {
-        if (!isNotTerminal()) {
+        if (isNotTerminal()) {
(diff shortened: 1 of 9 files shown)


src/main/java/games/cantstop/CantStopGameState.java
```diff
@@ -97,7 +98,10 @@ protected CantStopGameState _copy(int playerId) {
 
     @Override
     protected double _getHeuristicScore(int playerId) {
-        return getGameScore(playerId);
+        if (!isNotTerminal()) {
```
rdgain, line 101: Shouldn't this be reversed? not+not = yes. so if (terminal) return game score, else return result (which should always be 0)? Maybe that's intended, just a double-check.

rdgain (approved):

PR #226, opened 2022-11-09
Added extra logging for results in RoundRobinTournament

GAIGResearch:
- Results are now reported with a final ranking of players, sorted by points. 
- Added option to output this report to a separate file ((new parameter argument), besides console output. 
- In RRTournament, this report can be also set to happen after N games (new parameter argument), both to console and output file.


src/main/java/evaluation/ProgressiveLearner.java
```diff
@@ -196,7 +196,7 @@ private String injectAgentAttributes(String raw) {
     private void runGamesWithAgents() {
         // Run!
         RoundRobinTournament tournament = new RandomRRTournament(agents, gameToPlay, nPlayers,  true, matchups,
-                System.currentTimeMillis(), params);
+                finalMatchups, System.currentTimeMillis(), params);
```
hopshackle, line 199: this should be 'matchups', not 'finalMatchups'

GAIGResearch: Fixed!

PR #227, opened 2022-11-10
Tournaments draw points

GAIGResearch:


hopshackle (commented):
Two comments:
1) I would be in favour of not having a new drawScores argument for RoundRobinTournament, but just to make this new scoring the standard - it makes more sense to me.

2) The code should ideally also check for multiple WINs as well as DRAWS. Some games will report 2 (or more) players as both WINning. So the iteration over players should count how many WIN as well as DRAW.

GAIGResearch:
1. Sure, makes sense!
2. But wouldn't multiple WINs award 1 full point for each player? This would make sense to be cooperative/team games where all players in the team win the game, rather than competitive players that share a victory. For instance, a WIN in Pandemic would mean 1 point in the tournament for all players (they all couldn't have done better), while a DRAW in a competitive game like SushiGo would be less than 1 point to signify that they "failed" to win in a competitive game.

hopshackle:
I had forgotten about coop/team possibilities (which for the moment just means Pandemic...and possibly later Descent).

It does complicate things a little, and given the immediate requirement here is only for Sushi Go, then leave as is for my original point 2).

PR #229, opened 2022-12-12
Logistic regression

hopshackle:
This includes functionality to automatically learn heuristics from game trajectories using:
- Simple Ordinary Least Squares Regression (OLS)
- Support Vector Machines (SVM)
- Logistic Regression 

The entry point to use this is evaluation.ProgressiveLearner. 
- This starts with a specified starting heuristic (the default being NULL, which gives random play)
- This runs N games, and then learns a state evaluation heuristic from the trajectories of those games.
- On the next iteration an agent is added in to the growing set that uses the new heuristic.
- This stops after X iterations, and a round robin tournament is run between all the agents to estimate fitness

See documentation in the ProgressiveLearner class for further details.

This requires a State feature space to be defined (one of the inputs to ProgressiveLearner) that implements IStateFeatureVector. See examples in Dots and Boxes, Love Letter and Dominion.

The regression target can be any of WIN, ORDINAL, SCORE. (ORD_SCALE is also possible, which allows logistic regression to be used on the ORDINAL by converting it to a 0..1 range); otherwise Logistic Regression can only be used on a WIN target.

The new players.learners package contains the three OLS/SVM/Logistic learners that can be plugged in,


GAIGResearch:
All seems ok. I've added a new metric to GameMetric to record the name of the game (utilities/GameStatisticsListener doesn't exist anymore). To be merged into the metrics branch to make it compatible.

diegopliebana:
All seems ok. I've added a new metric to GameMetric to record the name of the game (utilities/GameStatisticsListener doesn't exist anymore). To be merged into the metrics branch to make it compatible.

diegopliebana (approved):
All seems ok. I've added a new metric to GameMetric to record the name of the game (utilities/GameStatisticsListener doesn't exist anymore). To be merged into the metrics branch to make it compatible.

src/main/java/core/Game.java
```diff
@@ -545,11 +547,18 @@ public final AbstractAction oneAction() {
                 action = observedActions.get(0);
                 currentPlayer.registerUpdatedObservation(observation);
             } else {
-                // Get action from player, and time it
-                s = System.nanoTime();
-                if (debug) System.out.printf("About to get action for player %d%n", gameState.getCurrentPlayer());
-                action = currentPlayer.getAction(observation, observedActions);
-                agentTime += (System.nanoTime() - s);
+                boolean explore = rnd.nextDouble() < exploreEpsilon;
```
rdgain, line 550: Why is this done as part of the game? I think it's complicating the game loop and I don't necessarily agree from a structural point of view that this should break the expectations of 'I give an action to the game and the game applies it'. If a different training procedure is needed, then I think this should be set up separately (even if it's probably not as easy...).

hopshackle: I agree - this was always terribly ugly.

I've now refactored this into AbstractPlayer which is a much better place for it. It can't really be any lower in the hierarchy as this could be used with any implementation.

src/main/java/core/AbstractPlayer.java
```diff
@@ -52,9 +58,20 @@ public void setForwardModel(AbstractForwardModel model) {
 
     /* Methods that should be implemented in subclass */
 
+    public final AbstractAction getActionWithExploration(AbstractGameState gameState, List<AbstractAction> possibleActions) {
```
rdgain, line 61: Small comment: this should be placed above the comment that says 'Methods that should be implemented in subclass' since it's final.

So this is changed so that this method is called by default from Game right? Can we call this instead 'getAction' to avoid confusion ('why are we getting actions with exploration all the time?'), and make the other one private '_getAction' if it's not accessed directly from outside anymore?

This function is also missing documentation, which would be helpful here for the future :)

hopshackle: Will add documentation!

I dithered about the naming convention - an went with getActionWithExploration() for two reasons:
1) To avoid having to change any existing agents, which all have to implement getAction().
2) And because this is always called from within the final method, so the name would be descriptive, but invisible to the casual implementor.

But 1) is resolvable with some adroit use of Intellij refactoring - so consider it done.

PR #230, opened 2023-01-05
Housekeeping

hopshackle:
Some end-of-year/twixtmas refactoring; mainly aimed to make it very slightly easier to implement new games.

1) Standardization of State/FM.reset()
Removed the requirement to implement a _reset() method on any extension of AbstractGameState. 
Currently games split their reset between GS and FM - which can lead to confusion (it has confused me in the past). 
Now the reset of the GS to its start of game status is the responsibility of FM.reset(), which has full access. 
This includes a single point for setting the initial game phase (which previously occured in two different places).

2) Default GamePhase
- Refactored to live in CoreContants.
- The Main phase is not used by all games, but was automatically set as the starting phase (unless explicitly overridden).
To reduce confusion I have refactored GS.setMainGamePhase() to be GS.setGaminPhase(...Main...) so it is clear what this does.

3) StandardForwardModel
This it intended to be the new standard to extend for new games. For the moment it just includes the functionality to call the action.execute() stuff which iis otherwise repeated in every single FM. An extension must then implement:
_afterAction(GA, Action);  which is clear when this is called.
This has been used for all games except for Pandemic (use of RuleBasedForwardModel) and ExplodingKittens (which is a very non-trivial re-write job to get it to fit).

4) Refactored TurnOrder.endRound
This now provides two hooks for sub-classes:
_endRound() for stuff that is genuinely at the end of a round; before we increment timers, publish the ROUND_OVER event and move the player on
_startRound() for set up at the start of the next round (which may not take place if the game ends)

All games then converted to this. The advantage is that the publication of ROUND_OVER events is now firmly in the framework for most games (and all new ones).
I also intended to do the same for endPlayerTurn() (and hence for TURN_OVER events); but ran out of time.

5) Refactored GS.gameEnd() code. 
A whole load of games implemented fundamentally the same logic once a game has finished - to decide who won and set playerResults. This is now standardised in AGS.endGame(). This works for all games with a Score and (optionally) a Tiebreak condition. This also standardises the use of WIN/LOSE/DRAW, which varied across games. (DRAW means two or more player in joint top position after considering tiebreaks).
It does not work for 2-player insta-win games (Connect4, TicTacToe), which have been left unchanged.

As part of this I also noticed that we have FM.endGame(), which is *only* called by Game (so not, for example, by MCTS or RHEA during rollouts). I suspect we could (should?) get rid of this to avoid confusion...thoughts?

6) Moved Game.terminateTimers to be before GAME_OVER event is published; otherwise incorrect data are reported.


src/main/java/core/AbstractGameState.java
```diff
@@ -106,41 +93,72 @@ void reset(long seed) {
         reset();
```
rdgain, line 93: Wasn't the reset() method on GS supposed to be removed?

src/main/java/core/AbstractGameState.java
```diff
@@ -106,41 +93,72 @@ void reset(long seed) {
...
     }
 
+    public final void setPlayerResult(Utils.GameResult result, int playerIdx) {
```
rdgain, line 96: Documentation issue: with methods shuffled around in the class (which is also really confusing Git and making it hard to tell what's actually changed/removed/added), the comments are no longer applying (and I'd like to keep them valid, so having getters / setters grouped, for example)

src/main/java/core/AbstractGameState.java
```diff
@@ -234,16 +258,33 @@ void checkActionsInProgress() {
         }
     }
 
+    public final void endGame() {
```
rdgain, line 261: This is functionality, changing the game state, I think it's better suited on the FM. I'd also say making this not final, so that games can process the end of game differently if needed.

src/main/java/core/StandardForwardModel.java
```diff
@@ -0,0 +1,24 @@
+package core;
+
+import core.actions.AbstractAction;
+import utilities.Utils;
+
+public abstract class StandardForwardModel extends AbstractForwardModel {
```
rdgain, line 6: I like this, a neat way to handle the multiple steps needed to be taken care of in next(). I'd propose expanding this even further, as there are 4 things done in most games in the next() function (that sometimes it's easy to forget about, so breaking it down like this and making it mandatory to implement is helpful):
- (done) Apply action
- (done) Execute any other required game rules post-action (e.g. change the phase of the game);
- Check for game end, and process the end of game appropriately
- Check if the player's turn or round needs to end, and do it

I'm not sure how much variety there is now in how these steps are done, but there are a lot of games that do these 4 steps pretty consistently so I think they can all be abstracted up like the first 2 you've done already.

rdgain: A follow-up thought: I think there's also a lack of consistency in the games implemented as to which of these things happen on the action execution, or the turn order itself, or in the next() in the FM. Would be good to look into this at some point and see if we can standardize things more...

src/main/java/core/turnorders/AlternatingTurnOrder.java
```diff
@@ -21,6 +21,15 @@ protected TurnOrder _copy() {
         return to;
     }
 
+    @Override
+    public void _endRound(AbstractGameState gameState) {
```
rdgain, line 25: I like these methods, but I don't like that all of these methods are implemented in the subclasses doing nothing. Should not be forced.

src/main/java/core/turnorders/TurnOrder.java
```diff
@@ -166,14 +167,25 @@ public void endRound(AbstractGameState gameState) {
 
         roundCounter++;
         if (nMaxRounds != -1 && roundCounter == nMaxRounds) {
-            gameState.setGameStatus(GAME_END);
```
rdgain, line 169: Related to comment of that functionality that is better on the FM. The initial idea was that this would just set the game_end flag in the game state, then the forward model would process the end of game appropriately.

src/main/java/games/catan/CatanGame.java
```diff
@@ -28,7 +28,7 @@
```
rdgain, line 28: Can we also remove the [x]Game classes for each game as part of this update? I feel like those confuse more too.

src/main/java/games/coltexpress/actions/roundcardevents/EndCardHostage.java
```diff
@@ -23,7 +23,6 @@ public boolean execute(AbstractGameState gs) {
         for (Integer playerID : locomotive.playersInsideCompartment){
             gameState.addLoot(playerID, new Loot(Purse, reward));
         }
-        gameState.setGameStatus(Utils.GameResult.GAME_END);
```
rdgain, line 26: Lost in the changes, is this set/checked somewhere else?

src/main/java/games/diamant/actions/OutOfCave.java
```diff
@@ -3,12 +3,14 @@
 import core.AbstractGameState;
 import core.actions.AbstractAction;
 import core.interfaces.IPrintable;
+import games.diamant.DiamantGameState;
 
 // This action is used when the player is out of the Cave
 public class OutOfCave extends AbstractAction implements IPrintable {
```
rdgain, line 9: Proposed refactor for Diamant: there are many action classes that do nothing (or do the exact same thing). Should be just one class, maybe with an enum parameter for the different options.

src/main/java/games/dicemonastery/DiceMonasteryTurnOrder.java
```diff
@@ -18,14 +18,15 @@
 public class DiceMonasteryTurnOrder extends TurnOrder {
 
     public DiceMonasteryTurnOrder(int nPlayers, DiceMonasteryParams params) {
-        super(nPlayers, params.YEARS);
+        super(nPlayers, params.YEARS * 4 - 1);
```
rdgain, line 21: I spot hard-coded numbers that I don't like

rdgain (changes requested):
I generally like these changes, I think it does help simplify things, thanks for the work! Just have a few comments for consistency in the changes and logic.

Something I was thinking of adding in the interest of making it easier to implement new games is to add a 'template game', that has all the core classes implemented returning empty things and having '//TODO's and explanatory comments. Importantly with links in the comments to the places where things need to be defined outside the game package (e.g. GameType and ParameterFactory). 

2 things I don't like when making a new game are
- that you have to jump around to various places to register your new game (and those are now also in 2 different classes which are also in different packages, with the separate parameter factory in 'core'). Proposed additional refactor: have a 'game register' package which breaks down the GameType class into clear places where you have to define your GS, FM, params and GUI classes. And also removing unused methods in there, like the string to game type method.
- and also that extending abstract classes creates methods that return null by default, so things crash when you try to run and the errors are not always intuitive (especially for the 'getAllComponents' error). I don't think there's a workaround for this, so having a 'template game' that returns empty things, 0s, 'this' etc. with todos and comments would be good. 

But I can do this last thing :)

src/main/java/core/AbstractGameState.java
rdgain: Afterthought: I'm getting a bit concerned of functionality split between FM and turn order. We have setup and endGame on FM, but the turn order has functionality for start round (which is often also done in setup) and end round, also calling end game in this format from the game state on max rounds reached. I think we need to restructure this a bit so it's more clear who does what...

src/main/java/core/AbstractGameState.java
hopshackle: No - it's the abstract _reset() method I've removed; I don't see anything wrong with the reset() method as it stands, as game implementors need know nothing about it.

hopshackle:
> 2 things I don't like when making a new game are
> 
> * that you have to jump around to various places to register your new game (and those are now also in 2 different classes which are also in different packages, with the separate parameter factory in 'core'). Proposed additional refactor: have a 'game register' package which breaks down the GameType class into clear places where you have to define your GS, FM, params and GUI classes. And also removing unused methods in there, like the string to game type method.
> * and also that extending abstract classes creates methods that return null by default, so things crash when you try to run and the errors are not always intuitive (especially for the 'getAllComponents' error). I don't think there's a workaround for this, so having a 'template game' that returns empty things, 0s, 'this' etc. with todos and comments would be good.
> 
> But I can do this last thing :)

I fully agree! Taking that first bullet point into scope is a great idea.

And ditto, a 'template game' could be helpful.

src/main/java/core/AbstractGameState.java
hopshackle: Not a problem - I'll reorganise.

src/main/java/core/AbstractGameState.java
hopshackle: "I think we need to restructure this a bit so it's more clear who does what..."

Yes - let's have that discussion. In this particular case putting endGame() on the GameState makes sense I think purely because the corresponding FM.endGame() is only used in one place (in Game); while code on the GameState to end the game is called directly all over the place. This is an issue in MCTS, RHEA. RMHC etc where currently the game is ended using GameState methods.

Very happy to move this to use an FM method - but it requires rather more refactoring.

src/main/java/core/StandardForwardModel.java
hopshackle: "A follow-up thought: I think there's also a lack of consistency in the games implemented as to which of these things happen on the action execution, or the turn order itself, or in the next() in the FM. Would be good to look into this at some point and see if we can standardize things more..."

You are not kidding! In a few cases I've just had to leave existing games untouched as converting them to a cleaner format woudl be too dangerous (especially in the absence of good unit test coverage).  Very happy to extend further - and this should also tie in with the discussion on turn order. 

My thinking general thinking is that for backwards compatibility we have a new model way of implementing a game, but if you want to you can make a spaghetti mess in the old style. (Unless we want to commit to the work to re-implement the games that are problematic.)

src/main/java/core/turnorders/AlternatingTurnOrder.java
hopshackle: Do you mean 'should be forced'?

I am also torn between making these abstract (and hence forcing an implementor to make a decision); and having a dummy implementation in some concrete parent class to avoid code boilerplate and stub methods that don't do anything and clutter up implementation classes.

src/main/java/core/turnorders/TurnOrder.java
hopshackle: "The initial idea was that this would just set the game_end flag in the game state, then the forward model would process the end of game appropriately."

I really don't like that idea. It seems to introduce the opportunity of bugs in the gap between setting the flag and actually implementing stuff. Are there any use cases where we can;t just end the game *immediately* it's over? To discuss.

src/main/java/games/catan/CatanGame.java
hopshackle: Yes...but I thought you liked them (for testing purposes without getting in the way of Game)

src/main/java/games/coltexpress/actions/roundcardevents/EndCardHostage.java
hopshackle: Yes - that is now done in endGame()

src/main/java/games/diamant/actions/OutOfCave.java
hopshackle: Agreed - that's how I'd do it - but it works, is straightforward to understand given how simple the game is, so changing it now feels a little too purist.

src/main/java/games/dicemonastery/DiceMonasteryTurnOrder.java
hopshackle: Oh alright then - I suppose I can add a parameter to say how many seasons there are in a year....just in case planetary cosmology changes.

src/main/java/games/diamant/actions/OutOfCave.java
rdgain: Actually, on this - I've done this refactor already with the Map Elites work we did a while back, that I still have to get into the main TAG repo. So scrap this comment, it'll be done... eventually.

src/main/java/games/dicemonastery/DiceMonasteryTurnOrder.java
rdgain: It can also be a comment to say what the 4 is :)

src/main/java/core/turnorders/AlternatingTurnOrder.java
rdgain: But this dummy implementation could just be on the abstract turn order, instead of cluttering all the other subclasses.

src/main/java/games/catan/CatanGame.java
rdgain: I don't :) - I think the easiest is to just use Game. With the exception of the games that do some extra stuff in their Game class, I think Pandemic had some custom competition setup in there.

PR #237, opened 2023-02-07
Metric updates

hopshackle:
Changes from Metric Merge

- Re-implemented IGameListener as interface
- Added ability to override default Events for metrics (to avoid need for separate metrics for GAME_OVER, ACTION_CHOSEN, END_ROUND when data is identical)
- RoundRobinTournament GameResult reporting now converted to Metrics system
- TurnOrder refactor had left both getTurnOwner() and getCurrentPlayer() methods on State - with some inconsistent usage. This has been tidied up, with getCurrentPlayer() used in most cases. The only difference is in what they return while an ExtendedActionSequence is in progress.
- Refactored the getTiebreak() functionality on AbstractGameState to be more robust to implementor error
- Refactored Rollout Termination in MCTS (needed after TurnOrder refactor, as this no longer worked properly for SelfOnly trees), and also added new unit/integration tests for this that surfaced bugs in Sushi Go and Love Letter left over from the TurnOrder refactor (bugs now fixed)
- ProgressiveLearner now works again
- Removed the ugly setExploration() method that Raluca rightly complained about from AbstractPlayer (used by ProgressiveLearner). This is now implemented with a Decorator pattern so it does not pollute the core class.

Update to processing of IExtendedSequences. (potentially more far-reaching)
- Corrected (I claim) to now register the action immediately *after* execution. Previously this was done before execution, which made little sense, and was (I think) due to the structure of ForwardModel. The new StandardForwardModel makes it more straightforward to register actions after execution, but before the _afterAction method is called to move on the game state. We only register the just-executed action with the top of the Stack (taking into account the possibility that this might be itself).


src/main/java/core/AbstractGameState.java
```diff
@@ -394,19 +392,26 @@ public final Stack<IExtendedSequence> getActionsInProgress() {
...
+    public double getTiebreak(int playerId, int tier) {
+        return Double.MAX_VALUE;
     }
 
+    /**
+     * This sets the number of tieBreak levels in a game.
+     * If we reach this level then we stop recursing.
+     * @return
+     */
+    public int getTiebreakLevels() {return 5;}
```
rdgain, line 413: Should this be 5 by default?

src/main/java/core/StandardForwardModel.java
```diff
@@ -20,6 +22,22 @@ protected void _next(AbstractGameState currentState, AbstractAction action) {
         } else {
             throw new AssertionError("No action selected by current player");
         }
+        // We then register the action with the top of the stack ... unless the top of the stack is this action
+        // in which case go to the next action
+        // We can't just register with all items in the Stack, as this may represent some complex dependency
+        // For example in Dominion where one can Throne Room a Throne Room, which then Thrones a Smithy
+        if (currentState.actionsInProgress.size() > 0) {
+            IExtendedSequence topOfStack = currentState.actionsInProgress.pop();
+            if (topOfStack != action) {
```
rdgain, line 31: Should this not be equals?

src/main/java/evaluation/listeners/IGameListener.java
```diff
@@ -0,0 +1,113 @@
...
+    Game getGame();
+
+    /**
+     * Create listener based on given class, logger and metrics class. TODO: more than 1 metrics class
+     * @param listenerClass - class of listener, full path (e.g. evaluation.metrics.MetricsGameListener)
+     * @param logger - class of logger (e.g. evaluation.loggers.SummaryLogger)
+     * @param metricsClass - class of metrics, full path (e.g. evaluation.metrics.GameMetrics)
+     * @return - GameListener instance to be attached to a game
+     */
+    static IGameListener createListener(String listenerClass, IStatisticLogger logger, String metricsClass) {
```
rdgain, line 43: This creates a metrics listener specifically. Should either rename method to be clear, or leave this static on the metrics listener class.

rdgain (approved):
Only a couple of comments.

src/main/java/core/StandardForwardModel.java
hopshackle: I put this as a reference level comparison deliberately - but on reflection the edge case I was thinking of is probably too extreme; and your approach is safer.

src/main/java/core/AbstractGameState.java
hopshackle: Five was the random number we mentioned when discussing this :-)

It should be more than enough for most purposes (so no need to override this method in new games); and won't have a performance impact if not implemented.

src/main/java/evaluation/listeners/IGameListener.java
hopshackle: Good point. Done.

PR #238, opened 2023-02-24
Love Letter Update

rdgain:
Changes outside of the package are mostly due to renaming WIN, LOSE, DRAW game results to *_GAME to allow distinguishing from *_ROUND in games which are played over multiple rounds where players may be eliminated before coming back for the next round.

- Fixed a bunch of bugs in Love Letter
- Simplified code
- Introduced abstract class PlayCard that all Love Letter actions subclass (and offloading a bunch of the common code, variables and checks onto this class instead). Executes the "move card from player hand to their discard pile", then the card's effect, if any, and executable (not executable if everyone targettable is protected, for example).
- Easier access to more game information for AI players / metrics system (and metrics updated too).
- All actions return correct copies and toString methods for display in the GUI
- GUI updated with a GUI listener for round change rather than the trickier check done previously

A thought on game result enums, currently often used to decide if a player should have actions generated for them or not. I see this more as a tree structure, i.e. 3 states are key, PLAYING (because game is ongoing, player is expected to make frequent decisions), PAUSE_PLAYING (if ROUND_WIN/LOSE, or just passed for the current round, but then coming back for more play later), and FINISHED_PLAYING (which can be due to WIN_GAME, LOSE_GAME, DRAW_GAME). I'm not sure if it's worth to rethink the structure of these a little for more clarity?


src/main/java/core/CoreConstants.java
```diff
@@ -57,9 +57,12 @@ public enum ComponentType {
     }
 
     public enum GameResult {
-        WIN(1),
-        DRAW(0),
-        LOSE(-1),
+        WIN_GAME(1),
```
hopshackle, line 60: Not sure GameResult is the right place to store WIN_ROUND statuses - but given LoveLetter is doing this anyway this is still an improvement!
However, I do think the values for any ROUND status should be 0.0, as the game is ongoing.

src/main/java/games/loveletter/actions/BaronAction.java
```diff
@@ -2,65 +2,63 @@
...
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 
-
 /**
  * The Baron lets two players compare their hand card. The player with the lesser valued card is removed from the game.
  */
-public class BaronAction extends core.actions.DrawCard implements IPrintable {
```
hopshackle, line 17: I'm getting errors when running ForwardModelTester on LoveLetter - and the problem seems to be in the Baron action. Will dig deeper.

src/main/java/games/loveletter/actions/PlayCard.java
```diff
@@ -0,0 +1,97 @@
...
+import games.loveletter.cards.LoveLetterCard;
+
+import java.util.Objects;
+
+public abstract class PlayCard extends AbstractAction {
+    final int playerID, targetPlayer;
+    final LoveLetterCard.CardType cardType, targetCardType, forcedCountessCardType;
+    final boolean canExecuteEffect;
+
+    public PlayCard(LoveLetterCard.CardType cardType, int playerID, int targetPlayer, LoveLetterCard.CardType targetCardType, LoveLetterCard.CardType forcedCountessCardType, boolean canExecuteEffect) {
```
hopshackle, line 17: Hurrah - much nicer then inheriting from DrawCard!

src/main/java/utilities/Utils.java
```diff
@@ -190,7 +190,7 @@ public static <T> Map<T, Double> normaliseMap(Map<T, ? extends Number> input) {
         return input.keySet().stream().collect(toMap(key -> key, key -> input.get(key).doubleValue() / sum));
     }
 
-    public static double range(double value, double min, double max) {
+    public static double clamp(double value, double min, double max) {
```
hopshackle, line 193: A better name.

PR #244, opened 2023-06-16
Puerto Rico and the death of GameReport

hopshackle:
This PR contains all the public changes made as part of the UK Games Expo project. This mainly includes:
- Puerto Rico implementation
- Updated metrics system (with further updates)

Plus some bug fixes and quiality of life improvements.

It also includes the integration of GameReport and RoundRobinTournament so that all of this functionality is available (with extra knobs on) from a single point of entry (RoundRobinTournament). See the updated help text in that class for details. This also now runs nicely with some default parameters.

A new data/listeners directory contains some useful standard listener JSON implementations (and MCTS and RHEA metrics are now moved out of the AbstractPlayer implementations, and into new Metrics classes)


src/main/java/core/OneShotExtendedAction.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Objects;
+import java.util.function.Function;
+
+/**
+ *  This is useful where we have one action that requires a single follow-on action by the same player that
+ *  we are 100% confident *must* be the next action taken in the game.
+ *
+ *  This is a wrapper around a function that computes the available actions for the next step in the sequence.
+ */
+public class OneShotExtendedAction extends AbstractAction implements IExtendedSequence {
```
rdgain, line 16: This class better in actions package?

src/main/java/core/Game.java
```diff
@@ -799,119 +790,6 @@ public String toString() {
         return gameType.toString();
     }
 
-    public void setupVideoRecording(String filename, String formatname,
```
rdgain, line 802: I'd prefer to leave the video recording in.

src/main/java/evaluation/tournaments/RoundRobinTournament.java
```diff
@@ -2,140 +2,131 @@
...
 import java.util.*;
 import java.util.stream.Collectors;
 
 import static core.CoreConstants.GameResult;
+import static evaluation.tournaments.RoundRobinTournament.TournamentMode.*;
+import static java.util.stream.Collectors.toList;
 import static utilities.Utils.getArg;
 
 
 public class RoundRobinTournament extends AbstractTournament {
```
rdgain, line 26: If this mashed with GameReport and GameReport is gone... can we have the main class to run things in the evaluation package (rather than evaluation.tournaments) and named so that it's clear it's *the* main thing to run for game/agent evaluation?

src/main/java/core/OneShotExtendedAction.java
hopshackle: Good call. I'll move.

src/main/java/core/Game.java
hopshackle: Looks like an accidental merge issue - I'll get it resurrected.

PR #245, opened 2023-06-20
PyTAG + Action Spaces

martinballa:
This is a large PR with various changes to the framework. The 2 main changes are coming from the ActionSpaces that allow altering the action space computation, the other big change is PyTAG which implements a Reinforcement Learning interface in JAVA which serves as the main entry point for RL agents.
PyTAG's main entry point is in ```core.PyTAG```
PyTAG requires the implementation of 2 interfaces
- ```IOrderedAction```: which generates a tree structure with the available actions in a fixed order
- ```IVectorisable```: extracts the state representation from the game state either as a vector or as a JSON object


src/main/java/core/Game.java
```diff
@@ -936,8 +952,19 @@ public static void main(String[] args) {
         players.add(new RandomPlayer());
         players.add(new RandomPlayer());
 //        players.add(new MCTSPlayer());
-//        MCTSParams params1 = new MCTSParams();
+        MCTSParams params1 = new MCTSParams();
+        params1.gatherTreeRecorder = true;
+        params1.discardStateAfterEachIteration = false;
```
hopshackle, line 957: All fine - but advance warning of an incoming refactor that removed all the EIDG code from MCTS directly, and moves it to a standalone MCTSListener that ties in with the Metrics system

src/main/java/core/Game.java
```diff
@@ -512,14 +520,21 @@ public final AbstractAction oneAction() {
 
         // Get actions for the player
         s = System.nanoTime();
-        List<AbstractAction> observedActions = forwardModel.computeAvailableActions(observation);
+        List<AbstractAction> observedActions = forwardModel.computeAvailableActions(observation, currentPlayer.getParameters().actionSpace);
```
hopshackle, line 523: I am not keen on the Game having a ForwardModel and using it to get available Actions, and then throwing these away so that the player uses their own Action space. 

As just discussed with Raluca, it would be cleaner if the Player strores the Forward Model / Action Space that it is using. The Game can then use this directly here to give a clean split of responsibility.

The important other thing is that the AbstractGameState does not have a reference (even via Parameters) to the ForwardModel or ActionSpace...these should be Player level parameters.

src/main/java/core/interfaces/IVectorisable.java
```diff
@@ -0,0 +1,20 @@
+package core.interfaces;
+
+import core.actions.AbstractAction;
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+
+import java.util.List;
+
+public interface IVectorisable {
+
```
hopshackle, line 10: I think there are two separate functions here, that should be two separate interfaces:
1) getJSON() - which is an interface implemented by a GameState if it provides a nice JSON representation.
2) the double[] arrays for use in RL. 

For 2) I suggest we use the existing IStateFeatureVector, and add the missing two methods (the normalised version can be a default?). This means that PyTag takes in an IStateFeatureVector as an additional argument with GameState, and uses this to generate the double[] feature representation. We can then plug-and-play and compare different state feature representations for the same game easily.

src/main/java/core/interfaces/IOrderedActionSpace.java
```diff
@@ -0,0 +1,14 @@
+package core.interfaces;
+
+import core.AbstractGameState;
+import utilities.ActionTreeNode;
+
+public interface IOrderedActionSpace {
```
hopshackle, line 6: Should this be ITreeActionSpace

src/main/java/evaluation/listeners/IGameListener.java
```diff
@@ -26,10 +31,9 @@ public interface IGameListener {
      * <p>
      * This is useful for Listeners that are just interested in aggregate data across many runs
      */
-    void allGamesFinished();
-
-    default boolean setOutputDirectory(String out, String time, String players) {return true;}
```
hopshackle, line 31: I think this (And other stuff) is from the slightly order metric system...just be aware that the newer version is now in master, so see what breaks when you merge that in,

src/main/java/games/loveletter/actions/deep/DeepBaronAction.java
```diff
@@ -0,0 +1,36 @@
+package games.loveletter.actions.deep;
+
+import core.AbstractGameState;
+import core.actions.AbstractAction;
+import games.loveletter.LoveLetterGameState;
+import games.loveletter.cards.LoveLetterCard;
+
+import java.util.List;
+
+/**
```
hopshackle, line 10: General comment that these Actions have no toString() method implemented - which is very helpful in debugging, and also in metrics output and on the GUI history panel...

src/main/java/evaluation/JsonRunner.java
```diff
@@ -0,0 +1,21 @@
+package evaluation;
+
+import core.interfaces.IGameRunner;
+import utilities.Utils;
+
+/**
+ * Receives a JSON config file as argument and runs the evaluation. Can run:
+ *  - Game Report {@link evaluation.GameReport} e.g. json/experiments/gamereport.json
+ *  - Tournaments {@link evaluation.tournaments.RoundRobinTournament} e.g. json/experiments/tournament.json
+ */
```
hopshackle, line 10: Very cool!

hopshackle (changes requested):

hopshackle:
Also - Catan fails the ForwardModel Tests (tests.games.fmtester), as copying the game state changes the hashcode (which it shouldn't):
Running Game 1 of Catan with seed -1280492308 at Wed Jun 21 17:45:00 BST 2023
Problem on state copy - orig/copy hashcodes are -433429348/-2028848153
	Orig: games.catan.CatanGameState@e62a649c
	Copy: games.catan.CatanGameState@87123be7

hopshackle:
The OMATests fail because chunks of the setup have been removed? Everything in test.players.mcts.OMATests

src/main/java/players/mcts/SingleTreeNode.java
```diff
@@ -557,13 +560,14 @@ protected void advance(AbstractGameState gs, AbstractAction act, boolean inRollo
...
         while (gs.getCurrentPlayer() != id && gs.isNotTerminalForPlayer(id) && !(inRollout && finishRollout(gs))) {
             //       AbstractGameState preGS = gs.copy();
             AbstractPlayer oppModel = opponentModels[gs.getCurrentPlayer()];
-            List<AbstractAction> availableActions = forwardModel.computeAvailableActions(gs);
+            List<AbstractAction> availableActions = forwardModel.computeAvailableActions(gs, params.actionSpace);
             if (availableActions.isEmpty())
-                throw new AssertionError("Should always have at least one action possible...");
-            AbstractAction action = oppModel.getAction(gs, availableActions);
+                throw new AssertionError("Should always have at least one action possible..." + (action != null? " Last action: " + action : ""));
+            action = oppModel.getAction(gs);
```
hopshackle, line 570: I'm slightly uncomfortable with this. 
First we calculate all the actions...and then ignore them when we get the actual action from the player (or oppModel). This is very inefficient.

We should just generate the agents once. In line with the proposals in the Game class; either we never generate the actions here, and rely on the Player doing this correctly; OR we get the ActionSpace/FM the player is using and generate the cations here (for error checking), before passing the list to the agent to make a decision, without recalculating everything.

src/main/java/players/mcts/SingleTreeNode.java
```diff
@@ -856,11 +860,13 @@ protected double[] rollout(double[] startingValues, int lastActor) {
...
             while (!finishRollout(rolloutState)) {
-                List<AbstractAction> availableActions = forwardModel.computeAvailableActions(rolloutState);
-                if (availableActions.isEmpty())
-                    throw new AssertionError("No actions available in rollout!");
-                AbstractAction next = opponentModels[rolloutState.getCurrentPlayer()].getAction(rolloutState, availableActions);
+                List<AbstractAction> availableActions = forwardModel.computeAvailableActions(rolloutState, params.actionSpace);
+                if (availableActions.isEmpty()) {
+                    throw new AssertionError("No actions available in rollout!" + (next != null? " Last action: " + next.toString() : ""));
+                }
+                next = opponentModels[rolloutState.getCurrentPlayer()].getAction(rolloutState);
```
hopshackle, line 869: As previous comment - we are duplicating effort by calculating all actions twice.

src/main/java/players/mcts/MultiTreeNode.java
```diff
@@ -124,11 +124,11 @@ public void oneSearchIteration() {
                 // note that different players will enter rollout at different times, which is why
                 // we cannot have a simple rollout() method as in SingleTree search
                 AbstractPlayer agent = opponentModels[currentActor];
-                List<AbstractAction> availableActions = forwardModel.computeAvailableActions(currentState);
+                List<AbstractAction> availableActions = forwardModel.computeAvailableActions(currentState, mctsPlayer.params.actionSpace);
                 if (availableActions.isEmpty())
                     throw new AssertionError("We should always have something to choose from");
```
hopshackle, line 130: And duplication of action calculation

hopshackle (changes requested):

src/main/java/core/components/BoardNode.java
```diff
@@ -131,7 +131,7 @@ public String toString() {
     }
 
     @Override
-    public final int hashCode() {
```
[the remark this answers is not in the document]
rdgain: It's precisely made not final :)

src/main/java/games/loveletter/actions/deep/DeepBaronAction.java
rdgain: They do on some super class, and I preferred this to more granular toStrings that end up confusing information more. In my tests they showed up properly in metrics / GUI

rdgain:
> Also - Catan fails the ForwardModel Tests (tests.games.fmtester), as copying the game state changes the hashcode (which it shouldn't): Running Game 1 of Catan with seed -1280492308 at Wed Jun 21 17:45:00 BST 2023 Problem on state copy - orig/copy hashcodes are -433429348/-2028848153 Orig: games.catan.CatanGameState@e62a649c Copy: games.catan.CatanGameState@87123be7

Fixed this, the random number generator slipped into the hashcode, without it it's fine. Not sure about the OMATests though

src/main/java/core/interfaces/IOrderedActionSpace.java
martinballa: renamed it to ITreeActionSpace

src/main/java/core/interfaces/IVectorisable.java
martinballa: Separated the feature extractor interfaces as suggested. Moved all the functions out of the GameStates and implemented them as separate classes for each game. PyTAG now has a "registry" that is used to check which games implement which interfaces and uses them to convert the game states with the right feature extractors.

hopshackle (approved):

PR #249, opened 2023-07-05
New game: 7 Wonders

rdgain:
By Arya Kakaroz (QMUL UG student project 22/23)

Note: the core of the game is implemented and playable. It's missing complex actions on player wonder boards, some yellow and purple cards, and cards that offer a choice of resource. To be completed up to the full game in the future.


src/main/java/core/StandardForwardModel.java
```diff
@@ -116,7 +116,7 @@ public void endPlayerTurn(AbstractGameState gs) {
      * @param gs                     - current game state.
      * @param firstPlayerOfNextRound the first player to act in the next round
      */
-    public final void endRound(AbstractGameState gs, int firstPlayerOfNextRound) {
+    public void endRound(AbstractGameState gs, int firstPlayerOfNextRound) {
```
hopshackle, line 119: I think we should keep endRound() as final. This is because it has the code that published END_ROUND events, which otherwise people will forget to include (this is why we made it final in the great TurnOrder refactor).

The ForwardModel can and should implement a specific set of end Round functionality, part of which is to call this method.

src/main/java/core/StandardForwardModel.java
```diff
@@ -144,7 +144,7 @@ public final void endRound(AbstractGameState gs, int firstPlayerOfNextRound) {
      * End a round, with no change to the firstPlayer
      * @param gs - game state
      */
-    public final void endRound(AbstractGameState gs) {
+    public void endRound(AbstractGameState gs) {
```
hopshackle, line 147: Also - I think we should keep as final.

src/main/java/games/wonders7/actions/BuyResourceR.java
```diff
@@ -0,0 +1,100 @@
...
+    // Player chooses card to play
+    public BuyResourceR(String cardName){
+        super();
+        this.cardName = cardName;
+
+    }
+
+
+    @Override
+    public boolean execute(AbstractGameState gameState) {
```
hopshackle, line 27: Ouch - a lot of code duplicated between BuyResourceR and BuyResourceL. Would it be straightforward to make L/R a parameter?

src/main/java/games/wonders7/cards/Wonder7Card.java
```diff
@@ -0,0 +1,307 @@
...
+            if (combinedResources.get(resource)== null){return false;}
+            if ((combinedResources.get(resource)) < constructionCost.get(resource)) { // Checks whether player's resource (after 'buying' resources) count can now afford the card
+                return false; // Player can't afford card with bought resources
+            }
+        }
+        return true;
+    }
+
+    // Checks if neighbour on the left can provide resources to build the structure
+    public boolean isPayableL(AbstractGameState gameState){
```
hopshackle, line 216: another case of much code duplication between R and L

hopshackle (changes requested):
Not too shabby! The only thing I feel strongly about is ensuring that endRound in StandardForwardModel is final.

src/main/java/core/StandardForwardModel.java
rdgain: But this is the same as endPlayerTurn, which creates the TURN_OVER events, and is similarly not final. Feeling strongly on making them all final? (People can also always forget to call the super methods in their own implementation anyway, maybe documentation is enough to prevent this - in which case allowing non-final methods would be better I think, as you'd have more direct access to the documentation for that method you're trying to write that reminds you to do all the right functionality in addition to your own)

hopshackle: Yes. I agree. Both should be final!

PR #251, opened 2023-07-27
Simple agent refactoring + Ecs7002p 2023-2024 preparations 

diegopliebana:
Changes:
 - Basic MCTS now has its own BasicMCTSParams class. This agent is also in its own package now.
 - The state evaluation heuristic in RMHC has been moved to RMHCParams, for consistency with MCTS and RHEA agents.
 - Added a few json files for the simpler agents to the json/players directory.
 - Also added a couple of default tournament JSON files to json/experiments.
 - Minor code simplifications/beautifications to ease the understanding of the code for ECS7002P (23/24)

### DIFF
--- a/json/players/osla.json
+++ b/json/players/osla.json
@@ -7,7 +7,5 @@
     		"BULLETS_PLAYER" : 0.5,
 		"BULLETS_ENEMY" : 0.0,
 		"LOOT" : 0.5
-	},
-	"budgetType" : "BUDGET_TIME",
-	"budget" : 40
+	}
 }
--- a/src/main/java/players/basicMCTS/BasicMCTSParams.java
+++ b/src/main/java/players/basicMCTS/BasicMCTSParams.java
@@ -1,0 +1,58 @@
+package players.basicMCTS;
+
+import core.AbstractGameState;
+import core.interfaces.IStateHeuristic;
+import players.PlayerParameters;
+
+import java.util.Arrays;
+
+
+public class BasicMCTSParams extends PlayerParameters {
+
+    public double K = Math.sqrt(2);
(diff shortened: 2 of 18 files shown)


json/players/osla.json
```diff
@@ -7,7 +7,5 @@
     		"BULLETS_PLAYER" : 0.5,
 		"BULLETS_ENEMY" : 0.0,
 		"LOOT" : 0.5
-	},
```
hopshackle, line 10: While you're at it, it would be good to remove the ColtExpress specific points here - the only line needed is the ScoreHeuristic

src/main/java/players/basicMCTS/BasicMCTSParams.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public BasicMCTSParams() {
+        this(System.currentTimeMillis());
+    }
+
+    public BasicMCTSParams(long seed) {
+        super(seed);
+        addTunableParameter("K", Math.sqrt(2), Arrays.asList(0.0, 0.1, 1.0, Math.sqrt(2), 3.0, 10.0));
+        addTunableParameter("rolloutLength", 10, Arrays.asList(0, 3, 10, 30, 100));
+        addTunableParameter("maxTreeDepth", 10, Arrays.asList(1, 3, 10, 30, 100));
```
hopshackle, line 26: I strongly recommend ensuring that the default in addTunableParameter matches the value in the main class. So set maxTreeDepth to 1000 in both.

This has caught me out before (and caused me to have to rerun 6 weeks worth of experiments becasue the default you see at first glance is not the one actually used if you fail to specify it in the json)

hopshackle (changes requested):
A couple of minor changes suggested - purely to reduce student confusion

PR #263, opened 2023-11-02
Random seeds and MCTSParams refactor

hopshackle:
**1) Random seed refactor**
- When running a game with a (fixed) random seed and MCTS agents with (fixed) random seeds, we now always get the same result. (Variious changes across many games)
- A new parameter has been added for RunGames of distinctRandomSeeds=n to specify how many seeds to use in a Tournament. Id this is set, then we will run all games n times - once with each random seed. This allows us to ask questions like, "If we run 1000 games for each of 100 distinct random seeds, ...."
- AbstractGameState has been refactored so that a Random() is held at this level, and used consistently across all sub-classes. It should no longer be necessary for game-implementers to do anything pther than use the AbstractGameState.getRnd() method....the framework deals with the rest. This removes a few commonly duplicated lines common to most games. GameTemplate updated with the details.

**2) PlayerParameters refactor**
MCTSPlayer now uses super().parameters (and does not have a local copy of MCTSParams any more).
Similarly for RHEA/RMHCPlayers.

**3) MCTSParams versus MCTSPlayer refactor**
rolloutPolicy, opponentModel, advantageFunction (actionHeuristic) and stateHeuristic are now moved from MCTSPlayer to MCTSParams. (per suggestion from Diego)
This also removes these from being recorded on every TreeNode.

**3) Tuned Agents**
Added json/players/gameSpecific directory with some tuned MCTS agents for most games in the framework

**4) Dice Monastery removed** on basis it is not currentrly active, and can be resurrected as needed. This also means that all the unit tests should now be expected to work.


json/players/gameSpecific/_Readme.txt
```diff
@@ -0,0 +1,5 @@
+These files are all for MCTS agents tuned to play the game with a budget of 40ms or so.
+
+They are not tuned over all possible MCTS parameters, but in most cases will play significantly better than the default MCTS agent.
+
+Where a game has noticeably different settings for different player counts or higher budgets, then these are inlcuded in the file name.
```
rdgain, line 5: included - typo

src/main/java/core/AbstractParameters.java
```diff
@@ -25,12 +25,14 @@ public abstract class AbstractParameters {
     long incrementMilestoneS = 0;
 
 
-    public AbstractParameters(long seed) {
```
rdgain, line 28: A question on this - if we do want to run with a specific random seed, should a second constructor specifying this not be included?

hopshackle: There is a public method setRandomSeed(long seed) which can be used in this use case.

PR #264, opened 2023-11-04
Fix firstPlayer and blinds in Poker

raymondcheung8:
Change first player to rotate clockwise after every round, instead of being permanently fixed on player 0
Change first player to be left of dealer preflop and to be small blind postflop
Change big blind to always be right of first player, instead of being right of small blind
Change small blind to always be right of big blind, instead of being first player

### DIFF
--- a/src/main/java/games/poker/PokerForwardModel.java
+++ b/src/main/java/games/poker/PokerForwardModel.java
@@ -46,6 +46,10 @@ public class PokerForwardModel extends StandardForwardModel {
         // Player 0 starts the game
         pgs.setFirstPlayer(0);
 
+        // Player to the right of first player is big blind
+        // Assuming player 0 starts the game
+        pgs.setBigId(pgs.getNPlayers() - 1);
+
         // Set up first round
         setupRound(pgs);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/games/poker/PokerForwardModel.java
```diff
@@ -164,6 +170,8 @@ protected void _afterAction(AbstractGameState gameState, AbstractAction action)
                     // Round is over
                     roundEnd(pgs);
                 }
+
+                pgs.resetTurnOwner();
```
hopshackle, line 174: You don't actually need to set up a new method for this - you should use the framework method (on StandardForwardModel) of endPlayerTurn(gameState, int newTurnOwner).
Or, if we just move on to the next clockwise player (who has not been knocked out), then endPlayerTurn(gameState) will do this as default behaviour.

hopshackle (commented):
Many thanks Ray!

We'd spotted the issue with first player not rotating (with a fix now in the randomSeeds branch which has a Pull Request open, and should be merged into master in the next couple of weeks); but not the issue with the small and big blinds.

What I'll do is merge this PR into the randomSeeds branch, and then do some light changes to regularise the use of endPlayerTurn / endRound framework methods before that one goes into master (see my comment on the PR).
 
James

PR #267, opened 2023-11-21
Put all changes from descent_refactor into main descent branch

L0RDR0B:


rdgain (approved):
Massive PR, I suggest simply merging and we can review the code from then on.

PR #268, opened 2023-12-04
Random seed fixes

hopshackle:
This has two main components:

1) The ability to provide a file of specific random seeds to RunGames (e.g. if we find a good one for a game set up and want to run 1000 games with the same starting cards)

2) Fixes for mode=random, mode=sequential and focuPlayer in RunGames. Due to a bug introduced in the main changes for RandomSeeds (Pull Request #263) running a purely random tournament used the same seed for every single game. (It was fine for the default mode=exhaustive). 
All now combinations of modes now tested properly.

### DIFF
--- a/src/main/java/evaluation/metrics/GameMetrics.java
+++ b/src/main/java/evaluation/metrics/GameMetrics.java
@@ -319,6 +319,53 @@ public class GameMetrics implements IMetricsCollection {
         }
     }
 
+
+    public static class Actions2 extends AbstractMetric {
+        public Actions2() {
+            super();
+        }
+
+        public Actions2(Event.GameEvent... args) {
+            super(args);
+        }
+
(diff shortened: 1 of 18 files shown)


src/main/java/evaluation/metrics/GameMetrics.java
```diff
@@ -319,6 +319,53 @@ public Map<String, Class<?>> getColumns(int nPlayersPerGame, Set<String> playerN
         }
     }
 
+
+    public static class Actions2 extends AbstractMetric {
```
rdgain, line 323: Can we call this something more informative than Actions2 / add brief doc for why it's here?

rdgain (changes requested):

rdgain (approved):

PR #270, opened 2023-12-19
Some Descent fixes

rdgain:
- Removed card component references in some actions
- Attack actions: all phases are iterated through, but there's not always a reaction available. Fixed to always allow ending a phase without taking another action (and this is handled correctly to move to next phase).
- Added some more descriptive strings for attack actions for debugging
- Removed dice references in TombleCopyDefence action that doesn't actually use the stored references, unnecessary
- TriggerAttributeTest does not need to set itself in progress again after another action is done.
- SwapOrder keeps action reference, should copy it properly
- RerollAttributeTest needs to remove the test that was done by a figure when executing, because tests have a check for only being allowed to be done once per turn. Figure stores attribute tests in a list now, removing the last one added when a reroll is triggered to allow the test to be done again.

This removes crashes related to attribute tests and attack actions.

Still to investigate: Jain Fairwood seems to take infinite turns, something is not triggered properly, potentially in some associate ability. To check if this happens with other heroes too.

### DIFF
--- a/src/main/java/core/Game.java
+++ b/src/main/java/core/Game.java
@@ -540,7 +540,7 @@ public class Game {
             throw new AssertionError("No actions available for player " + activePlayer
                     + (lastAction != null ? ". Last action: " + lastAction.getClass().getSimpleName() + " (" + lastAction + ")" : ". No actions in history")
                     + ". Actions in progress: " + actionsInProgress.size()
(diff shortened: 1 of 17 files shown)


L0RDR0B:
This seeems great.

I did have a fix for RerollAttributeTests but that felt a lot more 'hacky' than your solution, so I shall rectify that and implement your version instead.

Thanks again!

PR #272, opened 2024-01-03
Restricted Play 

hopshackle:
This provides a basic implementation of Restricted Play.

The key interface is IPlayerDecorator, which can be applied to a Player to filter out certain actions, restricting their action space.

DecoratedForwardModel then provides support to extend this into the search space of MCTS, so that the action restriction is followed during search. (With the option to restrict the action completely, or just for one's own actions while allowing the opponent models to use it.)

MCTSParameters has a new parameter of 'actionRestriction', which is an IPlayerDecorator implementation. A simple (and hopefully) useful example is provided with ActionClassDecorator, which filters out all actions of a specific class.

An example of the config in MCTSParams is then, which will disallow an use of 'VisitShop' (unless there are not other valid options):

```json
"class":"players.mcts.MCTSParams",
"actionRestriction": {
	"class" : "core.forwardModels.ActionClassDecorator",
	"args" : ["games.newgame.actions.VisitShop"]
}
```


rdgain:
I like this, I think a worry would be that this ties the filter to the player, and not the game. So if you want to filter out a core action in one game, but use it in another still, it'd get hacky to set up if running on multiple games. I'm not sure what a good way to go is though, I also don't like these files getting too verbose... but maybe an option to specify which game the filter applies to (by default to all)? I assume they'll be quite game-specific anyway, so it would make sense to have the game specified.

I also don't love that the decorated FM has a state, it extends AbstractForwardModel and has another AbstractForwardModel wrapped inside. It feels a bit overly complicated. Can we make the player id reference final, and the decorator list? Do we expect these to change on the same object throughout a run?

hopshackle:
_"a worry would be that this ties the filter to the player, and not the game. So if you want to filter out a core action in one game, but use it in another still, it'd get hacky to set up if running on multiple games."_ 

This is by design - we want to have four players, say, in a game, each with a different Restriction so we can see what the impact is. And yes, the actual implementations are often going to be game-specific - but that is I think inevitable as they are dependent on the action space design.

_"I also don't love that the decorated FM has a state, it extends AbstractForwardModel and has another AbstractForwardModel wrapped inside. It feels a bit overly complicated. Can we make the player id reference final, and the decorator list? Do we expect these to change on the same object throughout a run?"_

Decorated FM does not have state (assuming decorators don't). This was the neatest way I could find without interfering with the main StandardForwardModel and changing that contract. Everything made final now to clarify the stateless nature.

rdgain (approved):

PR #281, opened 2024-04-29
MCTS Tree reuse

hopshackle:
This implements re-use of the tree in MCTS; currently working for:
- SelfOnly
- OneTree
- MultiTree

MCGS yet to be done.

This is governed by a single new boolean flag MCTSParams.reuseTree, which defaults to false.

This also amends the history tracked in AbstractGameState to include the ID of each player as well as the action taken - this makes tracking through the old tree to find the current position much more robust.


src/main/java/core/AbstractGameState.java
```diff
@@ -53,7 +54,7 @@ public abstract class AbstractGameState {
     protected ElapsedCpuChessTimer[] playerTimer;
 
     // A record of all actions taken to reach this game state
-    private List<AbstractAction> history = new ArrayList<>();
+    private List<Pair<Integer, AbstractAction>> history = new ArrayList<>();
```
rdgain, line 57: Just one small comment, if we can update the comment for what this history is to indicate what's recorded for the pair (not immediately obvious).

rdgain (approved):

PR #287, opened 2024-07-23
Mastermind

cayenwang:
Implemented Mastermind


src/main/java/games/GameType.java
```diff
@@ -200,7 +206,15 @@ public enum GameType {
     ChineseCheckers(2, 6,
             Arrays.asList(Strategy, Abstract),
             Arrays.asList(GridMovement),
-            CCGameState.class, CCForwardModel.class, CCParameters.class, CCGUIManager.class);
+            CCGameState.class, CCForwardModel.class, CCParameters.class, CCGUIManager.class),
+    Mastermind(1,1,
+            Arrays.asList(Simple, Abstract, CodeBreaking),
+            Arrays.asList(),
+            MMGameState.class, MMForwardModel.class, MMParameters.class, MMGUIManager.class),
+    Cluedo(2,6,
```
hopshackle, line 214: The pull request should just contain Mastermind (and any core changes required to support it). Cluedo is still in development so should be kept in its own branch.

src/main/java/games/mastermind/MMGameState.java
```diff
@@ -0,0 +1,140 @@
...
+        return true;
+    }
+
+    @Override
+    protected double _getHeuristicScore(int playerId) {
+        return getGameScore(playerId);
+    }
+
+    @Override
+    public double getGameScore(int playerId) {
```
hopshackle, line 85: getGameScore should be reserved for an actual game score as per the rules of the game - which you can argue in Mastermind is equal to the number of rows unused. (Or, that there is no concept of score, in which case always return 0).

The code here is for a (very reasonable) heuristic.

hopshackle (changes requested):
Looks very good - main point is to remove the Cluedo code from this PR to avoid putting a half-finished game into the main repository.

cayenwang:
I've implemented the changes. Not sure if I need to re-mark this PR as 'ready for review'

PR #290, opened 2024-08-27
MinMax search

hopshackle:
Added agent (SearchPlayer) that implements MinMax search. 

This will go to a specified searchDepth, and this 'depth' can be measued in any one of three ways:
- ACTION, the number of indvidual actions
- MACRO_ACTION, an 'action' is instead assumed to be 'all actions taken consecutively by the same player'. This caters for extended actions neatly.
- TURN. The game Turn counter.

In a simple turn-taking game like Tic-Tac-Toe or Connect 4, these are all equivalent. Where we have extended actions or and/or multiple decisions on our 'Turn', then the last two may be more relevant...but at the cost of additional time.

A budget can also be specified as a timeout. If this is exceeded then the search stops and return the best action found so far (likely random).

The old OSLAPlayer has been removed and replaced with a wrapper around the new SearchPlayer.

### DIFF
--- a/src/main/java/core/Game.java
+++ b/src/main/java/core/Game.java
@@ -580,7 +580,7 @@ public class Game {
                 if (debug)
                     System.out.printf("Game: %2d Tick: %3d\t%s%n", gameState.getGameID(), getTick(), action.getString(gameState));
 
-                agentTime += (System.nanoTime() - s);
+                agentTime = (System.nanoTime() - s);
                 nDecisions++;
             }
             if (gameState.coreGameParameters.competitionMode && action != null && !observedActions.contains(action)) {
--- a/src/main/java/players/search/SearchPlayer.java
+++ b/src/main/java/players/search/SearchPlayer.java
@@ -8,6 +8,7 @@ import players.PlayerParameters;
 import players.simple.RandomPlayer;
 import utilities.Pair;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Random;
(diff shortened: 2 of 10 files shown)


src/main/java/core/Game.java
```diff
@@ -580,7 +580,7 @@ public final AbstractAction oneAction() {
                 if (debug)
                     System.out.printf("Game: %2d Tick: %3d\t%s%n", gameState.getGameID(), getTick(), action.getString(gameState));
 
-                agentTime += (System.nanoTime() - s);
+                agentTime = (System.nanoTime() - s);
```
rdgain, line 583: Was this not supposed to be an average? So summed to then be divided by ticks?

src/main/java/players/search/SearchPlayer.java
```diff
@@ -0,0 +1,114 @@
...
+import utilities.Pair;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Random;
+
+import static utilities.Utils.noise;
+
+public class SearchPlayer extends AbstractPlayer {
```
rdgain, line 14: Search is quite abstract, can we call it something more precise? Or at least add some docs to clarify what type of search this does

src/main/java/core/Game.java
hopshackle: There was not a common policy as nextTime and copyTime recorded the results for this particular action, while agentTime was a running total. This meant that the three of them were (very) inconsistent in the metrics.

This now makes them comparable when they are reported by the ComputationTimes metric (which is per ACTION).

rdgain (approved):

PR #291, opened 2024-08-27
War of the toads

hopshackle:
Mainly this is the implementation for War of the Toads, a 2-player simple card game by Alan Wallat, about to go to Kickstarter.

There are a number of changes/fixes added to support this:
1) Bug fix on PartialObservableDeck so that it overrides all the add methods of Deck. (Otherwise you get some unwanted behaviour on partial observability)

2) One-character fix to Metrics that was causing the last Action ever taken in a game to be ignored (or, more generally, the last data record)

3) Parameter Search and Class instantiation json files now have more robust support for JSON array formatting. 

3) MCTS parameter added for `instantiationClass`.
This allows you to specify the subclass of MCTSPlayer to actually be instantiated from the parameter definition. In Toads I've used this to do some useful jiggery-pokery with the hidden state before making a decision.

4) TunableStateFeatures (and related work on MCGS and IStateFeatureVector)
Improvements to definition of state definitions for MCGS (with a big improvement in performance using hash codes instead of Strings as the default)


src/main/java/core/AbstractPlayer.java
```diff
@@ -10,7 +10,7 @@
 public abstract class AbstractPlayer {
 
     // ID of this player, assigned by the game
-    int playerID;
+    protected int playerID;
```
rdgain, line 13: This change keeps creeping back :D but it should be package-private and not protected so that the game assigns it and players can only access it and not change it.

src/main/java/evaluation/listeners/MetricsGameListener.java
```diff
@@ -155,7 +155,7 @@ private String eventToIndexingColumn(IGameEvent e) {
         } else if (e == ROUND_OVER) {
             return "Round";
         } else if (e == TURN_OVER) {
-            return "Turn";
+            return "Tick";
```
rdgain, line 158: Why this change? This should be turn, no?

src/main/java/core/AbstractPlayer.java
hopshackle: Ooops. I was being sloppy to simplify my unit tests. I will go and complexify them.

src/main/java/evaluation/listeners/MetricsGameListener.java
hopshackle: Ah - this was subtle if I recall. The problem with using Turn is that it often resets at the end of each Round. This means that different TURN_OVER records overwrite each other...if we index them by Tick then this problem goes away, and we get a separate record for each Turn.

src/main/java/games/toads/abilities/Berserker.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.List;
+
+public class Berserker implements ToadAbility {
+
+
+    @Override
+    public List<Pair<Integer, BattleEffect>> tactics() {
+        // assassins copy their ally's tactics
+        return List.of(new Pair<>(5, (isAttacker, isFlank, br) -> {
```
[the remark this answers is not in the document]
hopshackle: You should have seen the earlier versions! ;-)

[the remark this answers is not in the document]
hopshackle: I've made some of these to be constants, but where this is a priority order of resolution that is very much trial and error, I can't see any meaningful way of making these constants. Happy to discuss.

src/main/java/evaluation/listeners/MetricsGameListener.java
rdgain: Mm that sounds hacky and odd when we have other concepts that are ticks. Should check this again

PR #299, opened 2024-11-08
Implement multithreading for NTBEA (ParameterSearch) and RoundRobinTournament (RunGames)

Joeytje50:
These changes implement a multithreading approach to both the ParameterSearch and the tournament classes. For tournaments (including those ran after completing ParameterSearch), this parallelizes the individual iterations evaluating the games, and for ParameterSearch this parallelizes the repeated runs, as specified in the `repeats` parameter. Unfortunately, due to being controlled by the `NTupleBanditEA` library, it is not possible to parallelize the individual game evaluations, which would make it possible to spread the workload of each evaluation much more efficiently.

In this PR there are also some changes to json/players/gameSpecific/TicTacToe.json, which were causing errors when this file was used anywhere (due to type mismatching).

I will annotate some parts of the code within Github, in order to make the review process easier. If there is some procedure generally used in PRs, please let me know, so I can make the process more efficient.


src/main/java/core/Game.java
```diff
@@ -495,13 +502,93 @@ public final void run() {
         }
     }
 
+    /**
+     * Runs an instance of the game, with the possibility to run fully parallel from any other game instances.
+     * Not to be used when there is a human player, only useful when parallelization needs to be possible.
+     * @return The final gameState object after finishing the game run(s)
+     */
+    public AbstractGameState runInstance(LinkedList<AbstractPlayer> players, int seed, boolean randomGameParameters) {
```
Joeytje50, line 510: This is essentially the same method as the `run()` method, except that it facilitates parallel execution. The reason this is a separate method, and not integrated into `run()` is because this function requires as few `synchronized` parts, in order to speed up execution, whereas `run()` requires a large synchronized block of code to facilitate GUI interaction such as pausing.
This meant I unfortunately needed to duplicate some of the code, although I removed all of the unnecessary pause/stop parts of the code. However, this also meant that most of the multithreading-specific code (such as increased number of scoped variables, instead of using class variables) is also limited to this runInstance method.

src/main/java/core/Game.java
```diff
@@ -495,13 +502,93 @@ public final void run() {
...
+            } else {
+                if (firstEnd) {
+                    if (gameState.coreGameParameters.verbose) {
+                        System.out.println("Ended");
+                    }
+                    terminate(gameState, forwardModel);
+                    firstEnd = false;
+                }
+            }
+            if (debug) System.out.println("Exiting synchronized block in Game");
```
Joeytje50, line 572: I kept this debug message in this version of the method, for consistency with debug output. If desired I could change this to something else, or remove it.

src/main/java/core/Game.java
```diff
@@ -495,13 +502,93 @@ public final void run() {
...
+     * Runs an instance of the game, with the possibility to run fully parallel from any other game instances.
+     * Not to be used when there is a human player, only useful when parallelization needs to be possible.
+     * @return The final gameState object after finishing the game run(s)
+     */
+    public AbstractGameState runInstance(LinkedList<AbstractPlayer> players, int seed, boolean randomGameParameters) {
+        AbstractGameState gameState = this.gameState.copy(); // our own copy of the gameState, to play games with
+        AbstractForwardModel forwardModel = this.forwardModel.copy(); // our own copy of the forwardModel, to avoid concurrency issues
+        reset(gameState, forwardModel, players, seed); // reset gameState before playing
+
+        synchronized (this) {
```
Joeytje50, line 515: Based on my testing, events posted to the listeners need to be executed sequentially. I did not do much research to the effect of these events, but based on my test runs, just adding a synchronized block to each of the event activations made everything work correctly.

src/main/java/core/Game.java
```diff
@@ -836,8 +937,7 @@ public static void main(String[] args) {
 
         /* Set up players for the game */
         ArrayList<AbstractPlayer> players = new ArrayList<>();
-        players.add(new RandomPlayer());
-        players.add(new RandomPlayer());
+        players.add(new BasicMCTSPlayer());
```
Joeytje50, line 940: I hope the listed players did not have any big significance to the master branch. If they are relevant, I'll just restore this to the Random+Random+MCTS state it used to be in.

json/players/gameSpecific/TicTacToe.json
```diff
@@ -2,11 +2,11 @@
   "budgetType": "BUDGET_TIME",
   "rolloutLength": 30,
   "opponentTreePolicy": "OneTree",
-  "MASTGamma": 0,
+  "MASTGamma": 0.0,
```
Joeytje50, line 5: Not relevant to this PR per se, except that MASTGamma and K are both of type double, meaning this json file was causing errors. I encountered this during testing of the multithreading, so fixed it here.

src/main/java/evaluation/optimisation/NTBEA.java
```diff
@@ -136,92 +135,55 @@ public void writeAgentJSON(int[] settings, String fileName) {
...
+                executor.submit(clone::runIteration);
+                clones.add(clone);
+            } else {
+                clone.runIteration();
+            }
         }
 
-        // After all runs are complete, if tournamentGames are specified, then we allow all the
-        // winners from each iteration to play in a tournament and pick the winner of this tournament
-        if (params.tournamentGames > 0 && winnersPerRun.get(0) instanceof AbstractPlayer) {
```
Joeytje50, line 148: This block and the next block of removed lines are all moved into the activateTournament() method, for legibility. This was done in a separate commit, to make it easier to follow all of the other canges made to this part of the code by looking at the other commits.

src/main/java/evaluation/tournaments/RoundRobinTournament.java
```diff
@@ -66,6 +71,7 @@ public RoundRobinTournament(List<? extends AbstractPlayer> agents, GameType game
                                 AbstractParameters gameParams, Map<RunArg, Object> config) {
         super(agents, gameToPlay, playersPerGame, gameParams);
         int nTeams = game.getGameState().getNTeams();
+        this.nThreads = (int) config.getOrDefault(RunArg.nThreads, 1);
```
Joeytje50, line 74: Optionally, I could make this also check for `Math.min(nThreads, iterations)`, but given that I find it unlikely for someone to specify 100 iterations and 128 threads, for example, I think this will be left as a responsibility to the person running the tournament to keep into account (the help string specifies that the number of iterations is parallelized, so I think it is reasonable to assume this won't be an issue). If more threads than iterations are specified, it'll just create a pool of threads that is larger than absolutely necessary, which is no big deal I figured.

src/main/java/evaluation/tournaments/RoundRobinTournament.java
```diff
@@ -235,7 +238,14 @@ public void createAndRunMatchUp(List<Integer> matchUp) {
                     List<Integer> matchup = new ArrayList<>(nTeams);
                     for (int j = 0; j < nTeams; j++)
                         matchup.add(idStream.getAsInt());
-                    evaluateMatchUp(matchup, 1, Collections.singletonList(seedRnd.nextInt()));
```
Joeytje50, line 238: Seeing this many repeated blocks of code checking for multi-threading, I'll push another commit to create a wrapper function to check if the call should be made to another thread, or in the same thread, to reduce duplicate code.

Joeytje50 (commented):
I've added annotations that are not intrinsically valuable to the code itself, but could be useful in the review of this PR. I'll also push another change as stated in one of my comments.

src/main/java/evaluation/tournaments/RoundRobinTournament.java
Joeytje50: Pushed; see `evaluateMatchUp(List<Integer> agentIDsInThisGame, int nGames, List<Integer> seeds, ExecutorService executor)`.

src/main/java/evaluation/optimisation/NTBEA.java
```diff
@@ -233,6 +195,102 @@ public Pair<Object, int[]> run() {
...
+            if (players.size() < nPlayers) {
+                // if we don't have enough players to fill the game, then we will need to use self-play
+                config.put(RunArg.mode, "exhaustiveSP");
+            } else {
+                config.put(RunArg.mode, "exhaustive");
+            }
+            config.put(byTeam, true);
+            config.put(RunArg.distinctRandomSeeds, 0);
+            config.put(RunArg.budget, params.budget);
+            config.put(RunArg.verbose, params.verbose);
```
Joeytje50, line 249: I've changed this to pass on the `verbose` run parameter. It seems to me this is intended behaviour, however if this is undesired I can change it back.

hopshackle: verbose should not be passed on to the final tournament (verbosity for NTBEA is not the same as verbosity for RoundRobinTournament)

hopshackle:
I've finally had a chance to look at this in more detail. The tl;dr is that I think (for reasons explained below) that parallelising RoundRobinTournament is a bit too dangerous; while parallelising NTBEA makes a lot of sense, but I am getting some errors on more complicated set ups.

**RoundRobinTournament**
- A better approach for the listener problem is to synchronise the Listeners (specifically MetricsListener and FeatureListener, plus their subclasses). This will mean that events are out of order in reporting files (or rather, are interleaved between multiple games running in parallel), but as long as the GameID is in the reporting file this is all fine. [I've committed changes for this to the PR branch]
- However, copying the game state within Game for each parallel run is going to cause problems (for very non-obvious reasons). The current undocumented design is that the Game.gameState is the master copy, and gameState.copy() does *not* copy over the attached listeners. This is to ensure that when using MCTS (or any other planning algorithm) the listeners do not pick up all the events from every planning iteration and rollout.  Hence, by copying the game state the link to the listeners is lost and we won't report on any events from the games. A quick fix would be to explicitly reattach all the listeners to the copied game state; but I suspect there will be other bugs nestling in this particular patch of undergrowth.
- No need to copy the forward model. Forward models should always be stateless.

**NTBEA**
- The whole thing falls over in a heap for me when I have search space files that reference other files (e.g. when tuning over coefficient files for feature based heuristics). I'm not at all sure why at the moment; it may be some thread unsafe libraries we're using. I need to spend some more time looking at this. This is not a problem with more 'vanilla' search spaces that just list possible settings directly.
- It may be slightly neater to take the CompletableFutures returned from the executor and generate the output from each individual run as soon as each is complete, rather than waiting for *all* iterations to finish. If I'm running 8 iterations on 4 threads, then I'd quite like to see the first set of output whiel the second set runs.

In summary:
- I'd rather avoid committing the parallelisation of RoundRobinTournament until we've had a lot more time to work out other implications given how single-threaded previous assumptions have been.
- NTBEA should be good; but I need to work out first where the deeper issue is on some of my set up.

Joeytje50:
> Hence, by copying the game state the link to the listeners is lost and we won't report on any events from the games. A quick fix would be to explicitly reattach all the listeners to the copied game state; but I suspect there will be other bugs nestling in this particular patch of undergrowth.
I believe when working on this, I had some issues with this as well. I attempted to fix this by adding a synchronized block around the calls to the listeners, but I must admit that I haven't really dug into the behaviour this caused. Perhaps a solution would be to clone the listeners and make them write to a separate place in memory/storage, and then aggregate all of the logging afterwards? That would probably require a bit more work than what I've done so far, but that way you would guarantee that the logging for different games won't get interwoven in the logs.

> No need to copy the forward model. Forward models should always be stateless.
I assumed as much as well, but I believe I was having some issues when I did not copy the forward model. It's been a short while since I did this, but I believe there were some errors that I attributed to concurrency issues, but I'm not entirely sure anymore. It could also be that I'm confusing this with something else, and that I only did some digging in the code, and concluded that I wasn't 100% sure forward models are always stateless, and therefore I might just have decided to copy it out of precaution. I'll have a further look at this at some point I guess.

src/main/java/evaluation/optimisation/NTBEA.java
hopshackle: That seems very fair.

PR #308, opened 2025-01-28
Implement rolloutIncrementType parameter for MCTSPlayer

Joeytje50:
Using this parameter, it is now possible to specify the rolloutDepth in terms of number of full turns, full rounds, or the old behaviour (which is still the default behaviour), being individual moves.


src/main/java/players/mcts/SingleTreeNode.java
```diff
@@ -932,7 +932,12 @@ protected boolean finishRollout(AbstractGameState rollerState) {
             return true;
         int currentActor = rollerState.getTurnOwner();
         int maxRollout = params.rolloutLengthPerPlayer ? params.rolloutLength * rollerState.getNPlayers() : params.rolloutLength;
-        if (root.actionsInRollout.size() >= maxRollout) {
+        int rolloutDepth = switch (params.rolloutIncrementType) {
+            case TICK -> root.actionsInRollout.size();
+            case TURN -> rollerState.getTurnCounter() - root.state.getTurnCounter();
```
hopshackle, line 937: The current convention is that the length of the rollout is defined from the point at which we leave the tree, not from the root node, which is what I think you have implemented here for TURN and ROUND.

there are two variables set up for the turn/round at the start of the rollout which can be used instead of root.state.getXXX
(also acceptable to add yet another parameter that enables the count to be from the root node if you want, just also make sure this applies to TICK as well) 

(And in passing I can see that there is a serious bug in (my) code for END_TURN and END_ROUND....oops...I'll fix that soon unless you fancied doing so :-) )

hopshackle (changes requested):
Looks good - I'd just request that you make the TURN/ROUND rollouts work on the same basis as TICK in terms of whether we start counting from the root of the tree or the leaf from which we start rollout. (The latter is the existing convention)

src/main/java/players/mcts/SingleTreeNode.java
Joeytje50: I've fixed this in my commit just now. I'm not sure what bug you mean in the END_TURN and END_ROUND code. Do you mean that it doesn't properly enforce that it doesn't correctly terminate only at the round/turn ends? If so, I could implement some fix for that, but I'm not quite sure what the most efficient approach to that would be. What I'm imagining would probably require some more complicated ways to track the actual end of turns/rounds, so if you've got plans to fix it efficiently I'll leave that to you.

hopshackle (approved):

PR #309, opened 2025-02-02
NTBEA improvements

hopshackle:
Four things included here:

1) Dependency on the old NTBEA libraries has been removed, and NTBEA code pulled into the main TAG repository in optimisation.ntbea package. This is a prerequisite to make the later improvements, and also gives us more flexibility in the future,

2) It is now possible to have a whole JSON Object as a moiety in an optimisation run. Previously one could only optimise over integers, doubles, strings and enums (via their string representation). Now the following JSON is legitimate when defining a search space:
```json
        "oppModelType" : "RANDOM",
        "heuristic" : [{
        	"class" : "players.heuristics.WinOnlyHeuristic"
        	},
        	{
        	"class" : "players.heuristics.ScoreHeuristic"
        	},        	
                {
        	"class" : "players.heuristics.NullHeuristic"
        	}
        ],
```

3) Added new parameter `evalsPerTrial` for ParameterSearch. This sets the number of games to run for each trial; the average of the results is then taken. This may reduce variance in some environments that outweighs the extra cost of running games [my personal experiments have usually suggested that it is better to just increase the number of trials].

3) Added ability to set the target of optimisation to be a specified quantile of output. This is irrelevant if optimising purely by win rate, but if optimising for score (or any non-binary outcome), then you may want to optimise for the median result, or for the best level that any given run has a 90% chance of reaching. This may help in provding robust recommendations.
This can now be achieved by setting `quantile: 50` or `quantile: 10` for these two examples in the Parameter Search config / command line. The default `quantile : 0` will use the mean outcome as the objective as now.


rdgain (commented):
Looks good! And approved to merge, but had some thoughts.

Something that would be useful is some documentation on using this version, maybe with some example templates for tuning game parameters too (I think there's something to that extent for a player) - some base class to run to optimize e.g. Tic Tac Toe parameters with some evaluator set up and an example JSON search space etc (all ready to run as demo), with nice comments for all the classes set up too.

In a similar vein, something that I would like too is an integration with the metrics system, maybe having some MetricEvaluator class that you could feed into usual parameters (opponents, number of games to repeat per config etc.) but also some metrics to maximize / minimize across all these games (which would also take care of attaching listeners and recording the data correctly). If that makes sense. I think this can be done within this version quite easily custom for an application, but can be left as future work - it's actually related to getting QD algorithms integrated too, so could be done as part of that at some point... :)

hopshackle:
There is some documentation on tuning a game at https://tabletopgames.ai/wiki/games/creating/game_tuning, to complement the documentation on tuning agents at https://tabletopgames.ai/wiki/agents/optimising

And I;ve just brought this up to date a little.

Not sure I follow completely on the Metrics integration. Let's discuss in person.

PR #313, opened 2025-02-19
Descent

hopshackle:
Integration of Descent into main code branch.
This includes:
- MCTS Forward Model tests working
- Conversion away from the old TurnOrder

GUI goes a bit wibbly after a few actions...but that wasn't me.

This is in preparation for MSc student to work on PCG of descent maps.


rdgain:
I'm unhappy with a few things in this, did a few commits to address some of the issues. Biggest thing is bringing game-specific components or imports into core. There was also at some point a change to Gridboard to handle board nodes exclusively like the graph board to make it more standardized, so redone this. Descent is unhappy now, though, and tests fail with null board nodes, to fix.

Also a couple other games fail some tests:
- Mastermind (all FM tests and TestMastermind, likely to do with the grid board change)
- Resistance (TestResistance) -> this is unrelated, unsure if the same fail happens on master too
- 7 Wonders (FM test with MCTS) -> also unrelated, unsure if the same fail happens on master too

rdgain (approved):

PR #321, opened 2025-05-05
Implementing Chess

Priwinn:
-Implements (western) chess
-Special moves: castling and en passant
-Includes edge case rules: Draw by repetition (can be disabled), Half move role
-Implements a drag and drop GUI


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "java.compile.nullAnalysis.mode": "disabled",
+    "java.configuration.updateBuildConfiguration": "interactive",
+    "java.debug.settings.onBuildFailureProceed": true
+}
```
hopshackle, line 5: You should exclude the .vscode directory from the repository (include in the .gitignore file). this is local to your setup only.

src/main/java/core/Game.java
```diff
@@ -845,34 +845,34 @@ public String toString() {
...
 
-        players.add(new HumanGUIPlayer(ac));
-        players.add(new HumanGUIPlayer(ac));
-        players.add(new HumanGUIPlayer(ac));
-//        players.add(new HumanConsolePlayer());
+        // players.add(new HumanGUIPlayer(ac));
+        // players.add(new HumanGUIPlayer(ac));
+        // players.add(new HumanGUIPlayer(ac));
+    //    players.add(new HumanConsolePlayer());
 //        players.add(new FirstActionPlayer());
```
hopshackle, line 877: Not a big deal, but it is better practice not to include any temporary testing changes as in Game.main() in the final PR.

src/main/java/games/chess/actions/EnPassant.java
```diff
@@ -0,0 +1,136 @@
...
+ * Include these parameters in the class constructor.</p>
+ * <p>They need to extend at a minimum the {@link AbstractAction} super class and implement the {@link AbstractAction#execute(AbstractGameState)} method.
+ * This is where the main functionality of the action should be inserted, which modifies the given game state appropriately (e.g. if the action is to play a card,
+ * then the card will be moved from the player's hand to the discard pile, and the card's effect will be applied).</p>
+ * <p>They also need to include {@link Object#equals(Object)} and {@link Object#hashCode()} methods.</p>
+ * <p>They <b>MUST NOT</b> keep references to game components. Instead, store the {@link Component#getComponentID()}
+ * in variables for any components that must be referenced in the action. Then, in the execute() function,
+ * use the {@link AbstractGameState#getComponentById(int)} function to retrieve the actual reference to the component,
+ * given your componentID.</p>
+ */
```
hopshackle, line 26: This boiler plate makes sense in the gametemplate package - but when you implement concrete classes you shoudl replace it with doc that makes sense to the implementation (i.e. here, one or two lines about En Passant...or even just delete it to avoid duplicated cruft)

src/main/java/games/chess/ChessGameState.java
```diff
@@ -0,0 +1,414 @@
...
+            return whitePieces;
+        } else if (playerId == 1) {
+            return blackPieces;
+        } else {
+            throw new IllegalArgumentException("Invalid player ID: " + playerId);
+        }
+    }
+
+    public void deletePiece(ChessPiece piece) {
+        if (piece == null) {
```
hopshackle, line 181: This is potentially dangerous as it can easily lead to bugs happening invisibly and being very difficult to track down. 
If code calls deletePiece() then they are expecting a piece to be present...if it isn't then a much better approach is to throw an Error (IllegalArgument, say).
This will highlight the bug as early as possible and make it much easier to fix

src/main/java/games/chess/actions/MovePiece.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public boolean execute(AbstractGameState ags) {
+        ChessGameState gs = (ChessGameState) ags;
+        ChessPiece piece = gs.getPiece(startX, startY);
+
+        //Delete the piece in the start position
+        gs.deletePiece(piece); // Remove the piece from its original position
+
```
hopshackle, line 31: What happens if there is no piece there? We should throw an error or do something?

src/main/java/games/chess/actions/Promotion.java
```diff
@@ -0,0 +1,79 @@
...
+        this.newPieceType = newPieceType;
+    }
+
+
+    @Override
+    public boolean execute(AbstractGameState gs) {
+        MovePiece move = new MovePiece(startX, startY, targetX, targetY);
+        if (!move.execute(gs)) {
+            return false;
+        }
```
hopshackle, line 42: In this case you should definitely throw an error - or better, do this within the move.execute() method. 
If the move fails then something unexpected has happened, and it is dangerous to let this fail silently and carry on as if everything was fine.

src/main/java/games/chess/components/ChessPiece.java
```diff
@@ -0,0 +1,137 @@
...
+        NOT_MOVED,
+        MOVED, 
+        NOT_RELEVANT
+    }
+    
+    //Wether the piece has moved or not (e.g. for castling, en passant, etc.), only relevant for pawns, kings and rooks
+    private MovedState movedState;
+
+    private int x;
+    private int y;
```
hopshackle, line 33: You are storing the position of a piece on the ChessPiece, and also recording the ChessPiece on an (x, y) position on the ChessBoard.

Both of these are individually very reasonable, but you have the same information stored in two separate places, which opens up all sorts of possible subtle bugs if the two sets of information get out of synch with each other.

This may be a refactor too far, but consider if there is a way to have a single source of the information. For example, could you make ChessPiece immutable and just have the mutable information stored on the chessboard.

src/main/java/games/chess/ChessGameState.java
```diff
@@ -0,0 +1,414 @@
...
+        List<Component> components = new ArrayList<>();
+        components.add(board);
+        return components;
+    }
+
+    @Override
+    protected ChessGameState _copy(int playerId) {
+        ChessGameState copy = new ChessGameState(getGameParameters(), getNPlayers());
+        copy.whitePieces = new ArrayList<>();
+        copy.whitePieces.addAll(whitePieces.stream().map(ChessPiece::copy).toList());
```
hopshackle, line 57: You could replace this I think with copy.whitePieces = new ArrayList(whitePieces); which is much more concise and avoids streams

src/main/java/games/chess/ChessGameState.java
```diff
@@ -0,0 +1,414 @@
...
+            return (playerScore - opponentScore)/100; // TODO: Normalize this value to be between 0 and 1
+        } else {
+            // The game finished, we can instead return the actual result of the game for the given player.
+            return getPlayerResults()[playerId].value;
+        }
+    }
+
+    @Override
+    public double getGameScore(int playerId) {
+        return 0;
```
hopshackle, line 132: This should check for a terminal game state, in which case the score is 0, 0.5, 1 for loss/draw/win?

src/main/java/games/chess/ChessParameters.java
```diff
@@ -0,0 +1,54 @@
...
+import core.AbstractParameters;
+import core.Game;
+import evaluation.optimisation.TunableParameters;
+import games.GameType;
+
+
+public class ChessParameters extends TunableParameters {
+
+    public int maxRounds = 300; // Maximum number of moves in the game before it ends in a draw
+    public int drawByRepetition = 3; // Number of times a position must be repeated for a draw by repetition to be declared, 0 for no limit
```
hopshackle, line 17: This is hiding a parameter of the same name in the grandparent AbstractParameters. 

You can use the existing functionality here - just set maxRounds to be the value required.

src/test/java/games/chess/ActionTests.java
```diff
@@ -0,0 +1,205 @@
+package games.chess;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.util.List;
```
hopshackle, line 5: Hurrah! Unit tests.

hopshackle (changes requested):

hopshackle:
I pressed submit on the PR review without summarising:
Main changes needed are:
1) Remove the private .vscode directory from the PR...this shoudl not go anywhere near the public repository.
2) Remove the duplicated comments that have been cut and paste from gametemplate (either replace them with something meaningful and useful for chess, or delete them to avoid cluttering the screen)
3) Throw errors when something unexpected happens; don't just fail silently. This is a quick route to very buggy and unreliable code.

src/main/java/games/chess/ChessParameters.java
Priwinn: Changing this to new name because using the existing functionality messes with the order of checks (checks maxRounds before checkmate)

Priwinn:
Made a bit of a mess rebasing to remove vscode and Game.java changes.
Added a couple of exception throws.
Removed remaining template text.
Minor fixes.

src/main/java/games/chess/components/ChessPiece.java
hopshackle: Perfectly reasonable conclusion.

src/main/java/games/chess/ChessGameState.java
hopshackle: Yes - my mistake...I forgot ChessPiece is mutable.

hopshackle:
Not sure if you're finished - but when I check out this branch, I have major compilation errors in ChessParameters, EnPassant and Castle due to unresolved merges?

Also - main remaining question is why have a separate maxChessRounds at all - just use the existing maxRounds?

Priwinn:
Sorry, not sure what happened but not it should compile correctly, also thanks to an anonymous playtester I found a bug for the bishops so that is fixed as well. I changed back to using the existing implementation of maxRounds.

hopshackle (approved):

PR #322, opened 2025-05-09
Exploding Kittens changes

longhousedev:
- Removed duplicate parameters
- Added a metric for chosen exploding kittens parameters

### DIFF
--- a/src/main/java/games/explodingkittens/metrics/ExplodingKittensMetrics.java
+++ b/src/main/java/games/explodingkittens/metrics/ExplodingKittensMetrics.java
@@ -1,0 +1,66 @@
+package games.explodingkittens.metrics;
+
+import core.interfaces.IGameEvent;
+import evaluation.listeners.MetricsGameListener;
+import evaluation.metrics.AbstractMetric;
+import evaluation.metrics.Event;
+import evaluation.metrics.IMetricsCollection;
+import games.explodingkittens.ExplodingKittensGameState;
+import games.explodingkittens.ExplodingKittensParameters;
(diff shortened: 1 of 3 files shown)


src/main/java/games/explodingkittens/metrics/ExplodingKittensMetrics.java
```diff
@@ -0,0 +1,66 @@
...
+            columns.put("SKIP_COUNT", Integer.class);
+            columns.put("FAVOR_COUNT", Integer.class);
+            columns.put("SHUFFLE_COUNT", Integer.class);
+            columns.put("SEETHEFUTURE_COUNT", Integer.class);
+            columns.put("TACOCAT_COUNT", Integer.class);
+            columns.put("MELONCAT_COUNT", Integer.class);
+            columns.put("BEARDCAT_COUNT", Integer.class);
+            columns.put("RAINBOWCAT_COUNT", Integer.class);
+            columns.put("FURRYCAT_COUNT", Integer.class);
+            return columns;
```
hopshackle, line 63: Why not use the same code (iteration over CardType.values()) here? That would also make the code robust to the addition of new cards if we implement EK expansions.

hopshackle (approved):
Looks good - one change suggested, but otherwise approved.

src/main/java/games/explodingkittens/metrics/ExplodingKittensMetrics.java
longhousedev: Good idea, will change

PR #326, opened 2025-05-23
Cant Stop Tunable Parameters

longhousedev:
Converted CantStopParameters from AbstractParameters to TunableParameters

Two parameters, DICE_NUMBER, and DICE_SIDES, require refactoring of the forward model before they can be tunable.


hopshackle (approved):
All OK - but note that changing the DICE won;t work - as four dice is effectively hard-coded in computeAvailableActions

PR #331, opened 2025-06-29
Auto features

hopshackle:
An implementation of Expert Iteration to provide a semi-automated way to combine tuning MCTS parameters with learning good(-ish) heuristic functions.

The new point of entry is ExpertIteration, which has been added to RunArg.

The basic inputs this takes are:
- a game / player count
- implementations of IStateFeatureVector and/or IActionFeatureVector for the game in question. These provide the raw inputs. The individual features can be ENUMs as well as numeric fields. At least one of these is needed (otherwise we have nothing to work on).
- NTBEA search space definitions for 'Value' and 'Action' phases of tuning. [See below]
- a basic starting agent to kick off the process (untuned MCTS for example)

The process then iterates over the following steps:
1) Run N games between the current agents [on the first iteration this will be just the starting agent]. Every A actions, consider what the current Oracle agent would do...the 'Oracle' is the current best agent with its budget multiplied by X. This generates data for the estimated value of the state, and the best action to take. This data is recorded for later training. The oracle recommendation is *not* taken in the actual game, but is only recorded for training purposes. [In a 'BASE' setting, no Oracle is used, and the actual decisions made and actual game outcomes are used in the nest step.]

2) Given the data from step 1), learn State and/or Action heuristics based on the Oracle data. This uses simple Least Square or Logistic regression [a Learner for each needs to be supplied]. These can aim for a number of different targets (e.g. just the Oracle's action, or the Advantage esimate of each action, or Visit Proportion defined by visit counts at the MCTS root node). This will not just use the features in the supplied FeatureVectors, but will also consider bucketing options on the numeric ones, and for ENUM features a sett of one-hot sub-features will be generated. Additionally interactions between all of these will be considered, with a basic greedy feature selection process and a Bayesian Information Criterion (tunable) level to stop overfitting.

3) A first phase of NTBEA tuning. This starts from the current best agent, with the new State heuristic inserted. MCTS (or other) parameters are then tuned to give a new agent.

4) A second phase of NTBEA tuning. This takes the agent from step 3) and now inserts the learned Action heuristic. A second NTBEA search space is now used to give the final new agent.

[At least one of steps 3 and 4 must be run, but it is perfectly possible to just use one of them.]

The new learned/tuned agent is added to the list of all agents, and we return to step 1).

The process ends once an agent wins three consecutive tournaments in a row, or one agent wins at least five. 

To support this there have been some more generically useful changes in the framework:
1) when loading objects from JSON, putting a file name (XXX.json) as the content of a field is now allowed, and this will automatically read in the file for the component. This helps reduce repetition.
2) when loading objects from JSON there is now an additional step/option c) below [a/b are existing]:
 a) Is the object class an instance of TunableParameters (if so, use that method of reading in parameters from key/value pairs)
 b) Does the JSON contain an args array, `args: ["arg1", 3.0, {...}, ... ]`, in which case look for a constructor with these arguments and use that.
 c) Does the class have a constructor that takes a JSONObject. If so, take the whole JSONObject and use this to construct the class. This allows much more flexibiliy than the ptrevious 2 approaches if you want to implement it.  
3) Some local areas of JSON parsing have been standardised by using the core functions in JSONUtils.
4) Heuristics can now have non-numeric features (they return a doubleVector or an objectVector). At least one is required.


L0RDR0B (changes requested):
Descent changes overall looks mostly fine?

Only major issue is getting rid of the TriggerAttributeTest class.

This was explicitly left to be an empty class to inherit from, such that when additional actions that caused Attribute Tests were involved, I could reuse the code from there rather than duplicating it across multiple cases.

As it stood, Howl was the only case for the tutorial scenario (as Barghests were the only Monsters spawned that could cause tests), but now that I have implemented further Monster abilities that use TriggerAttributeTests, this would complicate things a lot. There would be a major clash with what I currently have been implementing if this were approved.

L0RDR0B (approved):
Latest pushes in the main Descent branch include changes based upon these suggestions.

I've got it to work now on my end. Thanks again!

PR #336, opened 2025-09-17
Saboteur fixes

hopshackle:
Various fixes and refactors to Saboteur to fix some issues.

The main problem was that games were always ending in the same result depending on the random seed. This was because MCTS agents never got anywhere near a reward with search, and as a result the Saboteurs won every single round.

- Added parameterisation of the distance from the starting line to the gold nuggets
- Set boundaries on the expanse of the overall board to stop very silly placing of tiles (now the board ends at a line 2 squares to the left of the starting point, and also has a lmit of 5 squares above/below the top/bottom nuggets).
- Fixed the GUI so that you can now see the cards you have and the actions that are available.
- Also implemented the missing cards (Map and Rockfall cards were never added to player hands, and hence never used)

This revealed that there were some other deeper/more subtle problems. The key one was that the game did not allow a path from the start to track *through* an empty destination card, which made the game really very difficult for the miners! Hence I refactored parts of the route-checking to make this both more accurate and robust (and profiled to ensure this was not a performance issue).

### DIFF
--- a/json/players/gameSpecific/Saboteur/Saboteur.json
+++ b/json/players/gameSpecific/Saboteur/Saboteur.json
@@ -7,11 +7,10 @@
     "class": "players.heuristics.LeaderHeuristic"
   },
   "K": 1,
-  "exploreEpsilon": 0.1,
   "treePolicy": "UCB_Tuned",
   "MAST": "Both",
   "rolloutType": "MAST",
   "information": "Information_Set",
   "class": "players.mcts.MCTSParams",
-  "budget": 40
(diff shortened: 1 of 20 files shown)


rdgain (approved):
Nice updates and bug catches! Only comment would be nice to add to the parameters class a bit more documentation on how those parameters build the grid (and make the 2, +10/+5 from forward model setup lines 68-71 parameters in that set too, so all the info is in one place and it's clear what the grid dimensions would be and where the start / goal cards would be placed)

PR #342, opened 2025-10-27
PI-MCTS

vadithiyan:
- Added PISingleTreeNode
- Modified SingleTreeNode to handle PI-MCTS
- Modified MCTSEnums to include Perfect Inforamtion as an Information Policy and added A new enum for Perfect Information Aggregation Policy
- Added NumDeterminizations and PerfectInformationPolicy as Tunable parameters to MCTSParams
- Modified MCTSPlayer to use PISingleTreeNode when Perfect Information is selected as information policy
- Added PIMCTSMetrics class

### DIFF
--- a/src/main/java/players/mcts/MCTSPlayer.java
+++ b/src/main/java/players/mcts/MCTSPlayer.java
@@ -247,8 +247,15 @@ public class MCTSPlayer extends AbstractPlayer implements IAnyTimePlayer, IHasSt
         if (newRoot == null) {
             if (getParameters().opponentTreePolicy == MultiTree)
                 root = new MultiTreeNode(this, gameState, rnd);
-            else
-                root = SingleTreeNode.createRootNode(this, gameState, rnd, getFactory());
+            else {
+                if(getParameters().information == MCTSEnums.Information.Perfect_Information)
+                {
+                    root = new PISingleTreeNode(this, gameState, rnd);
+                }
+                else {
+                    root = SingleTreeNode.createRootNode(this, gameState, rnd, getFactory());
+                }
+            }
         } else {
             root = newRoot;
         }
--- a/src/main/java/players/mcts/MCTSPlayer.java
+++ b/src/main/java/players/mcts/MCTSPlayer.java
@@ -4,7 +4,6 @@ import core.AbstractForwardModel;
 import core.AbstractGameState;
 import core.AbstractPlayer;
 import core.actions.AbstractAction;
-import core.interfaces.IActionHeuristic;
 import evaluation.listeners.IGameListener;
 import core.interfaces.IStateHeuristic;
 import evaluation.metrics.Event;
@@ -14,8 +13,6 @@ import utilities.Pair;
 import utilities.Utils;
 
 import java.util.*;
-import java.util.function.BiFunction;
-import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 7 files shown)


src/main/java/players/mcts/PISingleTreeNode.java
```diff
@@ -0,0 +1,149 @@
...
+import core.AbstractGameState;
+import core.actions.AbstractAction;
+import players.PlayerConstants;
+
+import java.util.*;
+
+import static players.PlayerConstants.*;
+
+public class PISingleTreeNode extends SingleTreeNode {
+    SingleTreeNode[] roots;
```
hopshackle: A better name for this would be ForestNode (or something that indicates it constructs a number of different trees, and then aggregates results across them)

src/main/java/players/mcts/PISingleTreeNode.java
```diff
@@ -0,0 +1,149 @@
...
+    AbstractGameState state;
+    int numDeterminizations = 1;
+    MCTSPlayer mctsPlayer;
+    double epsilon = 1e-6;
+    //Action Stats variable to hold aggregated values
```
hopshackle: Use the value from MCTSParams

src/main/java/players/mcts/PISingleTreeNode.java
```diff
@@ -0,0 +1,149 @@
...
+    public void mctsSearch(long initialisationTime) {
+        initialiseRootMetrics();
+        initialisationTimeTaken = initialisationTime;
+        //Create Determinized Trees and Run MCTS on the Determinized Trees
+        for (int i = 0; i < params.numDeterminizations; i++) {
+            roots[i] = SingleTreeNode.createRootNode(mctsPlayer, state.copy(state.getCurrentPlayer()), rnd, mctsPlayer.getFactory());
+            roots[i].mctsSearch(initialisationTime);
+        }
+    }
+    //Returns the best action after search based on the chosen AggrergationPolicy
```
hopshackle: I reckon we can rely on the mctsSearch in the SingleTreeNode for this (and reduce code repetition a little)

src/main/java/players/mcts/MCTSPlayer.java
```diff
@@ -247,8 +247,15 @@ protected void createRootNode(AbstractGameState gameState) {
         if (newRoot == null) {
             if (getParameters().opponentTreePolicy == MultiTree)
                 root = new MultiTreeNode(this, gameState, rnd);
-            else
-                root = SingleTreeNode.createRootNode(this, gameState, rnd, getFactory());
+            else {
+                if(getParameters().information == MCTSEnums.Information.Perfect_Information)
```
hopshackle: Also, this doesn't take account of reuseTree (not the end of the world, but I wonder if we could implement this)

src/main/java/players/mcts/MCTSPlayer.java
```diff
@@ -247,8 +247,15 @@ protected void createRootNode(AbstractGameState gameState) {
...
+                {
+                    root = new PISingleTreeNode(this, gameState, rnd);
+                }
+                else {
```
hopshackle: A better form might be:
root = ForestNode.createRootNode(...) to keep the same pattern as for SingleTree node a couple of lines below [and then I can go an refactor MultiTreeNode to the same pattern]

hopshackle (changes requested):
Good - I;m being a bit fussy with some suggested changes.
The main one is that we should not have Perfect_Information as a new enum on Information, but just add a new parameter (num_determinizations) that extends the current Open_Loop / Closed_Loop information enums. 

A secondary point is some refactoring of PISingleTreeNode to change its name, and streamline bits of the code.

hopshackle (approved):

PR #343, opened 2025-11-26
Go fish

hopshackle:
New game - Go Fish


src/main/java/core/Game.java
dienn1, line 1: Revert changes in main()

dienn1 (changes requested):
Revert changes in main() in Game.java

PR #345, opened 2025-12-09
Power Grid Game added 

StephenHornish:
Contains fully implemented game Power Grid with a basic UI. In addition it has a basic leader Heuristic for MCTS and a Feature class for use in RL environments. The game matches the 2014 Deluxe Edition Power Grid North America Map except that there are no double cities such as NewYork and Mexico City.


src/main/java/core/PyTAG.java
```diff
@@ -367,86 +380,298 @@ public int sampleRNDAction(int[] mask, Random rnd){
...
+        } catch (Exception e) {
+            System.out.println("Exception during game initialisation: " + e);
+            e.printStackTrace();
+        }
+    }
 
-                    double reward = env.getReward();
-//                    System.out.println("at step " + steps + " the reward is " + reward + "player ID " + env.gameState.getCurrentPlayer());
+    
+ // --- ADDED FOR DEBUG DELETE LATER ---
```
hopshackle: PyTAG.java should not be left with any game-specific code in it (and I think this can all be safely removed). It would also be better to leave the example game in main() as TicTacToe as a better starting example.

hopshackle (changes requested):

src/main/java/games/powergrid/PowerGridGameState.java
```diff
@@ -0,0 +1,941 @@
...
+	    if (!Objects.equals(currentMarket, other.currentMarket)) return false;
+	    if (!Objects.equals(futureMarket, other.futureMarket)) return false;
+
+	    // Orders, sets, and lists
+	    if (!Objects.equals(turnOrder, other.turnOrder)) return false;
+	    if (!Objects.equals(roundOrder, other.roundOrder)) return false;
+	    if (!Objects.equals(bidOrder, other.bidOrder)) return false;
+	    if (!Objects.equals(plantsRan, other.plantsRan)) return false;
+	    if (!Objects.equals(RewardGiven, other.RewardGiven)) return false;
+	    if (!Objects.equals(income, other.income)) return false;
```
hopshackle: income is an Array, so this should use Arrays.equals

src/main/java/games/powergrid/PowerGridGameState.java
```diff
@@ -0,0 +1,941 @@
...
+        final PowerGridGamePhase phase = (PowerGridGamePhase) getGamePhase();
+        
+        final double wCities = 0.8;
+        final double wIncome = 1.2;
+        
+ 
+       //if the player wins at the end they get a reward else they get penalized 
+        if (this.getGameStatus() == CoreConstants.GameResult.GAME_END) {
+            CoreConstants.GameResult[] res = getPlayerResults();
+            CoreConstants.GameResult r = (res != null && playerId < res.length) ? res[playerId] : null;
```
hopshackle: The game score should always be the actual game score, there shoudl be no need to check for GAME_END. For Powergrid I would argue that the score is the number of powered citie; plus tiebreak of money.

(This differs to getHeuristicScore which can/should check for game end and normalise/clamp to a [-1, +1] range; and can use any number of heuristic estimates

src/main/java/games/powergrid/PowerGridHeuristic.java
```diff
@@ -0,0 +1,167 @@
...
+	 * @see PowerGridGameState
+	 * @see #getMaxExcluding(int[], int)
+	 * @see #shapedGap(double, double, double, double, double, double)
+	 */
+
+	@Override
+	public double evaluateState(AbstractGameState gs, int playerId) {
+	    PowerGridGameState s = (PowerGridGameState) gs;
+	    if (s.isGameOver()) {
+	        return (gs.getWinner() == playerId) ? 1.0 : -1.0;
```
hopshackle: This fails to compile - you want to use gs.getWinners(), which returns a Set of integers, allowing for draws.

(or actually better, check GameResult for the player)

hopshackle (changes requested):
Damn fine work if I may say so!

I've not gone through the detailed logic of the game, but have focues on reviewing the integration with the framework; and in particular common errors around deep copies and a few other conventions that are not 100% obvious on a first visit.

PR #348, opened 2025-12-23
RL custom reward strucutre 

StephenHornish:
Currently, PyTAG uses the TAG score mechanism directly as the reinforcement-learning reward signal. While this is sufficient for some environments, it limits the ability to define intermediate rewards or reward functions that are non-monotonic with respect to the final game score both of which are often necessary for long-horizon or phase-based games. To address this, I added to the default AbstractState class with a new method, getReward(). By default, this method simply returns the game score, ensuring full backward compatibility and preserving the behavior of all existing RL environments. With the introduction of getReward(), PyTAG now queries the state’s reward function rather than directly using the score. Developers may optionally override this method to define a custom reward structure that differs from the final scoring mechanism. In the case of Power Grid, this allows the score to remain consistent to the official game rules while enabling a separate reward signal (e.g., intermediate rewards during bureaucracy phases). As a result, reward design is cleanly decoupled from score computation, providing greater flexibility without breaking existing implementations.


StephenHornish:
Reopened this PR. It is cleanly mergeable and includes the custom getReward() reward hook plus the PowerGrid bug fix. Ready for review when convenient. If you do not want to merge it I can reclose it or leave it the way it is I can close it.

PR #350, opened 2026-01-05
Spades

hopshackle:
The game of Spades - now fully working.

### DIFF
--- a/src/main/java/games/spades/SpadesForwardModel.java
+++ b/src/main/java/games/spades/SpadesForwardModel.java
@@ -8,6 +8,7 @@ import core.components.Deck;
 import core.components.FrenchCard;
 import games.spades.actions.Bid;
 import games.spades.actions.PlayCard;
+import utilities.Pair;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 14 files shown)


src/main/java/games/spades/actions/PlayCard.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.util.AbstractMap;
+import java.util.Objects;
+
+/**
+ * Action representing playing a card in Spades.
+ */
+public class PlayCard extends AbstractAction implements IPrintable {
+    
+    public final FrenchCard card;
```
rdgain, line 19: This shouldn't store the component

src/main/java/games/spades/SpadesForwardModel.java
```diff
@@ -0,0 +1,318 @@
...
+import java.util.Map;
+
+public class SpadesForwardModel extends StandardForwardModel {
+
+    @Override
+    protected void _setup(AbstractGameState firstState) {
+        SpadesGameState state = (SpadesGameState) firstState;
+        Arrays.fill(state.teamScores, 0);
+        Arrays.fill(state.teamSandbags, 0);
+        for (int i = 0; i < 4; i++) {
```
rdgain, line 25: I guess this is used like this (throughout the game) assuming the player number is always 4, and there's some other assumptions based on this (e.g. teams). This is already imposed by the game type, but can we use e.g. number of players variable off the state instead (also for readability)? I'd like the possibility of experimenting with other player counts to see how the game breaks to be left open

src/main/java/games/spades/SpadesForwardModel.java
```diff
@@ -0,0 +1,318 @@
...
+    @Override
+    protected List<AbstractAction> _computeAvailableActions(AbstractGameState gameState) {
+        SpadesGameState state = (SpadesGameState) gameState;
+        List<AbstractAction> actions = new ArrayList<>();
+        int currentPlayer = state.getCurrentPlayer();
+        SpadesParameters params = (SpadesParameters) state.getGameParameters();
+
+        if (state.getGamePhase() == SpadesGameState.Phase.BIDDING) {
+            int team = state.getTeam(currentPlayer);
+            boolean nilAllowed = params.allowNilOverbid || state.getTeamScore(team) < 500;
```
rdgain, line 45: A few other odd numbers here, 500, 13 two lines down - params or documentation

src/main/java/games/spades/SpadesForwardModel.java
```diff
@@ -0,0 +1,318 @@
...
+            }
+        }
+
+        return winner;
+    }
+
+    /**
+     * Gets the value of a card for comparison (higher is better)
+     */
+    private int getCardValue(FrenchCard card) {
```
rdgain, line 176: The type in FrenchCard already has a number associated, which is the same in this case too, can just use card.type.number (or, if card.type == Number then return card.number).

src/main/java/games/spades/SpadesGameState.java
```diff
@@ -0,0 +1,310 @@
...
+        playerHands = new ArrayList<>();
+        tricksWon = new ArrayList<>();
+        playerBids = new int[4];
+        Arrays.fill(playerBids, -1); // -1 means not bid yet
+        tricksTaken = new int[4];
+        teamScores = new int[2];
+        teamSandbags = new int[2];
+        playerBlindNil = new boolean[4];
+
+        for (int i = 0; i < 4; i++) {
```
rdgain, line 54: I like components being created by the FM exclusively so as not to store any logic on the GS, but this is debatably acceptable init and cleared fine in the FM

src/main/java/games/spades/SpadesParameters.java
```diff
@@ -0,0 +1,59 @@
+package games.spades;
+
+import core.AbstractParameters;
+
+import java.util.Objects;
+
+public class SpadesParameters extends AbstractParameters {
+    
+    public final int winningScore = 500;
```
rdgain, line 9: These should not be final, we like changing parameters

rdgain (changes requested):

rdgain (approved):

PR #356, opened 2026-06-08
Simultaneous Move support

hopshackle:
This provides initial support in the core framework for Simultaneous Move games. This enables new games to have simultaneous moves, or for old ones with internal implementations (SushiGo, War of theToads, Diamant, Resistance...) to be converted to the new framework. However, it is backwards compatible with all existing games, so no conversion is required.

The advantage of converting a game is that SFP algorithms may then be able to take advantage of this.

AbstractGameState
- added new method to getSimultaneousPlayers(), that returns a List of the players who make a move at the same time
- a new method for redeterminise(int playerId) has been added to enable a clean separation between copy() and redeterminise(); which really are two different things. [see note at the end of this PR for plans here]

AbstractForwardModel
- added new methods to specify (optionally) the player for whom moves should be calculated in computeAvailableActions

IExtendedSequence
- added support for delegation from FM/State for simultaneous moves

Game
- amended the core loop to *assume* simultaneous moves, and use the new methods in AbstractGameState and AbstractForwardModel above
- default implementations of these new methods in AbstractGameState and AbstractForwardModel ensure backwards compatibility and the default assumption of sequential player actions

At the moment, only SushiGo has been converted to the new set up.

The split of _copy() and redeterminise() was required for SushiGo because the redeterminisation has to set the turnOwner, and _copy() is called at the start of copy(), before copy() then sets up the fields defined in AbstractGameState (overriding turnOwner).

This split into _copy() and redeterminise(in playerId) enables the latter to be called at the end of the copy() method, giving it the option of shuffling/obscuring data at the AGS level, which is theoretically desirable. It also provides a clean semantic split between, "the code that copies the state data", and "the code that shuffles this to account for imperfect information", which is in any case a better design that may reduce errors. 
The intention is therefore in a separate PR to fully implement this _copy/redeterminise split across all existing games...but not in this one.


src/main/java/core/Game.java
```diff
@@ -359,142 +348,135 @@ public final boolean isHumanToMove() {
...
                 s = System.nanoTime();
                 if (debug)
-                    System.out.printf("About to get action for player %d%n", gameState.getCurrentPlayer());
+                    System.out.printf("About to get action for player %d%n", activePlayer);
                 action = currentPlayer.getAction(observation, observedActions);
-                if (!observedActions.contains(action)) {
-                    throw new AssertionError("Action played that was not in the list of available actions: " + action);
-                }
-
+                if (actionValidation && !observedActions.contains(action))
```
rdgain, line 426: Not a big thing, but I'm wondering if in some games this may be allowed (so part of core games parameters instead, for example). Say, everyone throw a card down, if someone doesn't play a legal action, they're out of the game, or lose a point, but the rest of the actions are still applied. No game comes to mind that would use this mechanism, but I think there would be cases where the legal actions should still be applied while ignoring or penalizing the others, so I'd be in favour of keeping this definable at game level. Could be actionValidation = true by default with the behaviour being a function called on the FM for handling invalid actions (I think there may already be something like that, or at least it was at some point before on the FM), with the possibility of overriding this by individual games for different behaviours.

src/main/java/games/sushigo/SGForwardModel.java
```diff
@@ -134,6 +139,128 @@ protected void _afterAction(AbstractGameState currentState, AbstractAction actio
         }
     }
 
+
+
+    */
```
rdgain, line 144: Odd comments, class should be nicely documented especially as a first example of using proper simultaneous actions

src/main/java/games/sushigo/SGForwardModel.java
```diff
@@ -200,12 +327,12 @@ void revealCards(SGGameState gs) {
                 }
             }
         }
-        int expectedPlayerCards = gs.getPlayerHands().get(0).getSize();
-        for (int i = 1; i < gs.getNPlayers(); i++) {
-            if (gs.getPlayerHands().get(i).getSize() != expectedPlayerCards) {
-                throw new AssertionError("Player " + i + " has " + gs.getPlayerHands().get(i).getSize() + " cards, expected " + expectedPlayerCards);
-            }
-        }
+//       int expectedPlayerCards = gs.getPlayerHands().get(0).getSize();
```
rdgain, line 330: Should be cleaned up if no longer necessary

src/main/java/games/sushigo/SGForwardModel.java
rdgain: Any old code should also be removed to allow proper diff and reduce confusion

rdgain (changes requested):

PR #360, opened 2026-07-21
Slim PyTAG jar + release automation

martinballa:
## Summary
- Adds a second `maven-assembly-plugin` execution producing `target/TAG-pytag.jar` (~13MB vs ~404MB for the full `TAG.jar`), built from `src/assembly/pytag-jar.xml`. It keeps the core engine, all games, built-in players, and the Swing GUI/replay path, but excludes the Spark/Hadoop ML-learner, tablesaw analytics, and langchain4j/google-cloud dependency trees that [PyTAG](https://github.com/martinballa/PyTAG) never touches. Verified end-to-end against the real PyTAG package via JPype (vector + JSON obs examples, multi-agent).
- Fixes a crash (`NoSuchElementException`) in `SSPPlayerView`, `SGPlayerView`, and `GoFishPlayerView` when constructing the GUI with no human player (e.g. all-AI games, or PyTAG's RL agents) - they called `human.stream().toList().getFirst()` instead of falling back to `-1` like `LoveLetterPlayerView` already does. Found while sweeping games to verify the slim jar renders correctly.
- Adds `.github/workflows/release-pytag-jar.yml`: on publishing a GitHub release, builds the project and uploads `target/TAG-pytag.jar` as an asset named `TAG.jar`, matching the URL shape (`.../releases/latest/download/TAG.jar`) PyTAG's `jar_setup.py` expects.

## Test plan
- [x] `mvn package` builds both `TAG.jar` and `TAG-pytag.jar` cleanly
- [x] `TAG-pytag.jar` verified against a real `pip install -e .` PyTAG checkout (JPype), running `pt-action-masking.py` (vector obs, action masking) and `ma-random.py` (JSON obs, multi-agent) to completion
- [x] Swept ~11 games (incl. asset-heavy ones: Pandemic, ColtExpress, Dominion, Uno, Catan) launching the GUI manager on the slim jar with no exceptions
- [ ] Release workflow itself is untested against a real GitHub release (needs a maintainer to publish a release to trigger it)

### DIFF
--- a/.github/workflows/release-pytag-jar.yml
+++ b/.github/workflows/release-pytag-jar.yml
@@ -1,0 +1,36 @@
+name: Attach PyTAG jar to release
+
+# Runs when a maintainer publishes a GitHub Release (as has always been done for
+# this repo's v1.0/v2.0/... tags). Builds the project and attaches the slim
+# PyTAG runtime jar (target/TAG-pytag.jar, see src/assembly/pytag-jar.xml) to
+# that same release as an asset named TAG.jar - the exact filename and URL
+# shape (.../releases/latest/download/TAG.jar) that PyTAG's jar_setup.py
+# expects (github.com/martinballa/PyTAG).
+on:
+  release:
+    types: [published]
+
+permissions:
+  contents: write
+
+jobs:
+  build-and-upload:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


hopshackle (approved):
All looks good!

